Issue #2, opened 2026-03-21
feat: expose full VEP annotation flags + fjall streaming backend


mwiewior:
## Summary

Update `vepyr.annotate()` to expose all VEP annotation flags supported by the upstream `annotate_vep()` engine in `datafusion-bio-functions`, and add support for the new fjall streaming backend (`use_fjall` option).

## Background

The upstream `datafusion-bio-functions` now supports:
1. **Fully streaming contig-by-contig annotation** (PR #47) with per-contig memory reclamation
2. **Fjall KV backend** (PR #48) for variation lookup + SIFT — plugs into the same streaming pipeline via `"use_fjall": true` in `options_json`

The current `vepyr.annotate()` exposes only a subset of flags. This issue tracks exposing the full set.

### 1. Bump `datafusion-bio-functions` dependency

**File:** `Cargo.toml`

Update the pinned commit from `f22c24e` (pre-streaming) to the latest on `master` after PR #47 and #48 are merged. This brings in:
- `ContigAnnotationExec` streaming pipeline
- `PartitionedParquetCache` auto-detection
- `use_fjall` option support
- `KvCacheTableProvider::from_store()` constructor

Also update `datafusion` and `arrow` versions to match (currently DataFusion 52.1.0, Arrow 57.x in upstream).

### 2. Add `kv-cache` feature flag

**File:** `Cargo.toml`

The fjall backend requires the `kv-cache` feature on `datafusion-bio-function-vep`:
```toml
[features]
default = []
kv-cache = ["datafusion-bio-function-vep/kv-cache"]
```

### 3. Expose missing annotation flags in `annotate()`

**File:** `src/vepyr/__init__.py`

The current `annotate()` signature:
```python
def annotate(vcf, cache_dir, *, chrom=None, backend="parquet",
             extended_probes=False, everything=False, hgvs=False,
             shift_hgvs=False, reference_fasta=None, cache_size_mb=1024)
```

Should be expanded to:
```python
def annotate(
    vcf: str,
    cache_dir: str,
    *,
    chrom: str | None = None,
    backend: str = "parquet",
    # Variation lookup backend
    use_fjall: bool = False,
    # Annotation feature flags
    everything: bool = False,
    hgvs: bool = False,
    hgvsc: bool = False,
    hgvsp: bool = False,
    shift_hgvs: bool | None = None,  # None = auto (True when hgvs enabled)
    no_escape: bool = False,
    remove_hgvsp_version: bool = False,
[...]

PR #1, opened 2026-03-18
Bootstrap PyO3/maturin project with upstream Rust deps

mwiewior:
## Summary
- Scaffolded Python+Rust project with `uv init --lib --build-backend maturin`
- Configured Cargo.toml with pyo3 0.25, arrow 56 (pyarrow interop), datafusion 50.3, tokio
- Added both upstream biodatageeks crates as pinned git deps: `datafusion-bio-function-vep` and `datafusion-bio-format-ensembl-cache`
- Added CLAUDE.md with project overview, build commands, and architecture notes
- Verified: `cargo check` compiles all ~200 crates, `uv sync` builds the extension, Python import works

## Test plan
- [x] `cargo check` passes
- [x] `uv sync` builds and installs the extension
- [x] `uv run python -c "from vepyr import hello; print(hello())"` returns expected output

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #27, opened 2026-07-06
feat: Plugin cache

mwiewior:


mwiewior:
@claude pls review

mwiewior:
@codex pls review

mwiewior:
## Addressed review (Claude bot + Codex) → `36e293d`

**1. Temp clone + worktree leak (Claude #1).** `_resolve_plugin_manifest` is now a `@contextmanager`: on exit it runs `git worktree remove --force` (deleting the worktree dir *and* its registration in the source repo's `.git/worktrees/`) and `shutil.rmtree`s the online temp clone. A caller-supplied `plugins_repo` is left untouched — only the worktree we added is removed. `build_plugin_cache` now wraps the build in `with _resolve_plugin_manifest(...)`.

**2. Multi-source manifest silently partially overridden (Claude #2 / Codex P2).** `build_plugin_cache` (Rust) now rejects manifests with more than one `[[source]]` with a clear error — the public API takes a single `source_path` and can't map multi-part sources, so it fails fast instead of reading a stale placeholder path. (Extending to a per-part path map can come with the first real multi-source plugin.)

**3. `overwrite=false` not enforced (Codex P2).** Confirmed the v0.14.0 builder's `with_overwrite` is a no-op, so I added the guard in the wrapper: without `overwrite=True`, `build_plugin_cache` refuses to clobber an existing cache (checks `<root>/plugin/<name>/manifest.json`) instead of silently rewriting shards + manifest.

**4. `plugin_cache_root` on the streaming path (Claude #3 — confirmed wired).** `options_json` is built (including `plugin_cache_root`) *before* the `output_vcf` branch, and the LazyFrame path forwards the same blob to `_create_annotator` — so both output modes carry it. Added a test (`test_plugin_cache_root_reaches_streaming_options`) asserting this on the `output_vcf=None` path, alongside the existing VCF-output test.

### Verification
- New/updated tests: worktree-cleanup assertions (worktree removed, caller repo + registration left clean), multi-source rejection, overwrite guard, streaming-path `plugin_cache_root`.
- **906 passed, 2 skipped**; `cargo fmt` + `clippy -D warnings` + `ruff` clean. Rust guards verified locally via `maturin develop`.

@codex review
/cc @claude — re-review please

mwiewior:
## Addressed Codex P3 → `e19b120`

**Temp clone leaks on `git clone` failure.** Correct — the online `mkdtemp` + `git clone` ran *before* the `try/finally`, so a failed clone (bad url / network / credentials) left a stale `vepyr-plugins-*` dir. Moved the clone inside the `try` scope with `repo`/`worktree` initialized to `None`, and guarded the `finally` on what was actually created — so a partial/failed clone is cleaned up too, and the worktree-removal is skipped when no worktree was added.

New test `test_failed_clone_leaves_no_temp_dir` drives a failing clone (nonexistent `repo_url`) and asserts no `vepyr-plugins-*` temp dir is left behind. All 7 plugin-cache tests pass; `ruff` clean.

@codex review

mwiewior:
## Addressed Codex re-review → `83289a9`

**P2 — overwrite guard blocked incremental builds (my regression).** Correct. My guard fired on mere cache existence, so `chroms=["chr1"]` then `chroms=["chr2"]` (default `overwrite=False`) failed on the second call even though no chr2 shard existed. Since the builder **upserts** filtered builds, the guard now only fires for an **unfiltered** build (`chroms=None`) — which rewrites every chrom and truly clobbers the cache. Filtered `chroms=[...]` builds pass through and upsert, so incremental per-chromosome builds work. New test `test_filtered_build_not_blocked_by_overwrite_guard` confirms a filtered build reaches the builder (fails only on the missing variation cache, not the guard); the full-rebuild refusal test remains.

**P3 — docs.** Reworded the `[[source]]` section in `docs/plugins.md`: the cache format reserves `part`-suffixed multi-source blocks, but vepyr's public `build_plugin_cache()` takes a single `source_path` and rejects manifests with >1 `[[source]]` (not yet supported) — so users aren't pointed at a path that raises `ValueError`.

8 plugin-cache tests pass; `cargo fmt` + `clippy` + `ruff` clean. Rust change verified via `maturin develop`.

@codex review

mwiewior:
## Addressed Codex re-review → `860cba4`

**P2 — `chroms=[]` bypassed the overwrite guard.** Correct edge case. `chroms.is_none()` is false for `Some([])`, but `PluginCacheBuilder::with_chrom_filter` treats an empty list as *no filter* and rebuilds every shard — so an empty list could perform exactly the full-cache clobber the guard refuses. The guard now treats **None or an empty list** as a full build (`chroms.as_ref().is_none_or(|c| c.is_empty())`). New test `test_empty_chroms_treated_as_full_rebuild` covers it.

9 plugin-cache tests pass; `cargo fmt` + `clippy` + `ruff` clean. Verified via `maturin develop`.

@codex review

mwiewior:
@codex pls review

mwiewior:
@codex pls review

mwiewior:
@claude pls review

PR #40, opened 2026-07-29
test(e2e): release-116 5-plugin profiles + CSQ comparator fix

LucasJezap:
## Summary
- Add `merged_116_base` / `merged_116_5plugins` profiles to `run_annotation_fast.py`, comparing release 116 against a golden VEP output (`--everything --hgvs`, ClinVar+SpliceAI+CADD+AlphaMissense+dbNSFP together).
- Fix `compare_vcfs()`: pair CSQ entries by `Feature` (transcript ID) instead of zipping by index after independently sorting each side. The old approach silently misaligned every entry after the first transcript-set divergence between vepyr and VEP, reporting unrelated transcripts' field values as mismatches.
- Add `values_equivalent()`: tolerate float formatting differences (Rust's shortest-round-trip vs Perl's fixed-decimal printf, e.g. `"0"` vs `"0.00"`, `"0.57985"` vs `"0.579850"`) and VEP's `.` missing-value marker vs vepyr's empty string, instead of requiring byte-identical strings.

## Result (chr22, 50,861 variants, ~150k CSQ entries)
Before → after these two comparator fixes:
- Plugin fields: went from a confusing mix of false positives (CADD_RAW 89.7%, SpliceAI `DS_*` ~50%, dbNSFP/AlphaMissense CSQ-order noise) to **37/38 fields at exactly 100.0%**.
- Core fields (no-plugin base profile): went from ~40 fields showing apparent mismatches down to **84/86 at 100.0%**.
- The two real residuals that survive the fix are both investigated and documented, not comparator artifacts:
  - `Consequence`/`Feature_type`: 12 mismatches, all `vepyr: TF_binding_site_variant` vs `vep: intergenic_variant` — a real, narrow motif/regulatory-feature disagreement, unrelated to plugins.
  - `CADD_RAW`/`CADD_PHRED`: 3 mismatches, all the same variant (`chr22:29489578`, an 18bp insertion). Verified: the raw CADD source has the record, vepyr's cache key is correct, and VEP itself only emits the value on 1 of the variant's 4 overlapping transcripts (CADD is a per-variant, not per-transcript, plugin — this is VEP's own internal inconsistency between transcripts, not a vepyr defect).

## Caution
`Cargo.toml` carries a `[patch]` override pointing `datafusion-bio-function-vep` at a local sibling checkout (`../datafusion-bio-functions`), used to test against this session's local branch changes ahead of a release. It's path-relative and will break a build for anyone without that exact sibling checkout — please don't merge this as-is without either removing the patch or confirming the target `datafusion-bio-functions` tag already has what's needed.

## Test plan
- [x] Ran both new profiles on chr22 against the release-116 golden VEP reference, confirmed field-match rates above.
- [ ] Full-genome run (chr1-22) not yet done for these two profiles.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1193,7 +1193,6 @@ dependencies = [
 [[package]]
 name = "datafusion-bio-function-vep"
 version = "0.12.0"
-source = "git+https://github.com/biodatageeks/datafusion-bio-functions.git?tag=v0.14.0#86bd907fc31917351b263681fe103250a5a773ee"
 dependencies = [
  "ahash",
  "async-trait",
@@ -5351,7 +5350,7 @@ checksum = "accd4ea62f7bb7a82fe23066fb0957d48ef677f6eeb8215f372f52e48bb32426"
 
 [[package]]
 name = "vepyr"
-version = "0.1.1"
+version = "0.2.0"
 dependencies = [
  "arrow",
  "datafusion",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -46,3 +46,11 @@ serde_json = "1"
 datafusion-bio-function-vep = { git = "https://github.com/biodatageeks/datafusion-bio-functions.git", tag = "v0.14.0", features = ["cache-builder"] }
 datafusion-bio-format-ensembl-cache = { git = "https://github.com/biodatageeks/datafusion-bio-formats.git", tag = "v1.8.8" }
 datafusion-bio-format-vcf = { git = "https://github.com/biodatageeks/datafusion-bio-formats.git", tag = "v1.8.8" }
+
+# TEMP dev override (branch: plugin-test-fixes): pulls in the join.rs
+# tier-probe DISTINCT fix (dedup variation probe before the tier LEFT JOIN)
+# from a local checkout ahead of it landing on a released tag. Remove this
+# patch once datafusion-bio-functions cuts a release with the fix and bump
+# the `tag` above instead.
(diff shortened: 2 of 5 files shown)


mwiewior:
Superseded by #48, which ports this onto current `master`.

`master` moved to 7fcaf7a and aa4fd01 split `run_annotation_fast.py` into the `comparison/` package, so all five files here conflict. In the port:

- **The `Feature`-pairing fix is already on master, in a stronger form** — `_pair_entry_groups()` pairs on `(ALLELE_NUM, Allele, Feature)` and reports one-sided entries, so a transcript-set divergence can't cascade. Nothing to carry over.
- **`values_equivalent()` is ported**, but as exact float equality after parsing rather than `rel_tol=1e-4 / abs_tol=1e-6` — that `abs_tol` would call a real gnomAD AF of 6e-7 equal to zero, and every case documented here (`"0"`/`"0.00"`, `"0.57985"`/`"0.579850"`) is exactly equal once parsed anyway. It's also counted in its own `FormatOnly` column so the parity gate stays byte-strict.
- **The two profiles are ported** as `merged_plugins` / `merged_plugins_base`, with paths derived from `$DATA` instead of hardcoded.
- **The `[patch]` override is not ported**, per the caution in this description.

LucasJezap:
It sounds like this one is unnecessary now and could be closed, is that right @mwiewior ?

LucasJezap:
closing as it's probably done in the other PR

PR #45, opened 2026-08-23
feat(e2e): MD5 concordance comparator for VEP vs vepyr VCFs

mwiewior:
Adds `e2e-testing/scripts/md5_concordance.py` — an MD5-based concordance check between Ensembl VEP and vepyr VCF output, so parity can be asserted with a digest instead of a bespoke diff each time.

## Two modes

**`canonical`** (default) normalizes the serialization differences that are known to be cosmetic before hashing — QUAL rendered numerically, INFO and FORMAT keys sorted, FORMAT keys missing in every sample dropped. Equal digests mean equal annotation content.

**`strict`** hashes record bytes as-is. This is the parity target: it passes only once vepyr reproduces VEP's serialization exactly.

Header and body are hashed separately, because VEP appends run-specific provenance — `##VEP=` carries a wall-clock timestamp, an absolute cache path, and per-module git hashes — that can never match byte for byte. Those two lines are excluded from the header digest.

`--explain` classifies *how* mismatching records differ, so a failure names the cause rather than just reporting a digest mismatch.

Exit status: `0` concordance, `1` mismatch, `2` input error — usable as a CI gate.

## Usage

```bash
# sweep a release directory
./md5_concordance.py --results-dir ../results/116

# one pair, with a breakdown of what differs
./md5_concordance.py --pair vep_chr21.vcf vepyr_chr21.vcf --explain

# gate on exact bytes
./md5_concordance.py --results-dir ../results/116 --mode strict
```

## Verification

Run against `e2e-testing/results/116` (VEP 116.0 merged cache, GRCh38, HG002 v4.2.1, `--everything --hgvs --merged`):

- **canonical: PASS 22/22 chromosomes over 4,096,123 records**, exit 0. Digests were cross-checked against an independently written canonicalizer and match exactly (chr1 `68f3fb0f…`, chr6 `94487e34…`, chr21 `f1f1cade…`, chr22 `9392f95c…`).
- **strict: fails as expected**, and `--explain` reports exactly the four known serialization classes with no CSQ or value differences:

```
      20,000  QUAL format
      20,000  FORMAT KEYS (-['PS'] +[])
      20,000  SAMPLE1 keys
      19,925  INFO order
```

The header diff surfaces the seven lines vepyr currently drops (`##FILTER=<ID=PASS>`, `##fileDate`, `##CL`, both `##bcftools_norm*`) and the contig `assembly=` attribute.

## Notes

- Reuses `comparison.vcfio.open_text`, so plain, gzip, and BGZF inputs all work.
- Contig labels sort naturally (chr2 before chr10, chrX last).
- `--explain` counts the `PS` drop under both `FORMAT KEYS` and `SAMPLE1 keys`; both columns genuinely differ, so the count is accurate if slightly redundant.
- Pre-commit `ruff check` and `ruff format` both pass.

### DIFF
--- a/e2e-testing/scripts/md5_concordance.py
+++ b/e2e-testing/scripts/md5_concordance.py
@@ -1,0 +1,457 @@
+#!/usr/bin/env python3
+"""MD5 concordance between Ensembl VEP and vepyr VCF output.
+
+Two modes, same digest pipeline:
+
+  canonical (default)
+      Normalizes the serialization differences that are known to be cosmetic
+      before hashing: QUAL rendered numerically, INFO and FORMAT keys sorted,
+      FORMAT keys that are missing in every sample dropped. Two files with the
+      same canonical digest carry identical annotation content and differ only
+      in how they were written out.
+
+  strict
+      Hashes the record bytes as-is. This is the eventual parity target: it
+      passes only once vepyr reproduces VEP's serialization exactly.
+
+Headers are hashed separately from the body in both modes, because VEP appends
+run-specific provenance (``##VEP=`` carries a timestamp) that never matches and
+is not meant to.
+
+Exit status is 0 when every pair concords in the selected mode, 1 otherwise.
+
+Examples
+--------
+Compare one directory of per-chromosome runs::
+
+    ./md5_concordance.py --results-dir ../results/116
+
+Compare a single pair, and explain what differs::
+
+    ./md5_concordance.py --pair vep_chr21.vcf vepyr_chr21.vcf --explain
+
+Gate a release on exact bytes::
+
+    ./md5_concordance.py --results-dir ../results/116 --mode strict
+"""
+
+from __future__ import annotations
+
+import argparse
+from collections import Counter
+from dataclasses import dataclass, field
+import hashlib
+import os
+from pathlib import Path
+import re
+import sys
+
+sys.path.insert(0, os.path.dirname(os.path.abspath(__file__)))
+
+from comparison import vcfio  # noqa: E402
+
+# VEP appends these to the input header; they carry run-specific provenance
+# (wall-clock time, absolute cache path, per-module git hashes), so they can
+# never match byte for byte and are excluded from the header digest.
+VOLATILE_HEADER_PREFIXES = ("##VEP=", "##VEP-command-line=")
+
+MISSING = (".", "")
+
+# Default per-chromosome layout under e2e-testing/results/<release>/:
+#   fast_chr21/vep_chr21_merged.vcf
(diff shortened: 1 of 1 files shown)


mwiewior:
@claude pls review

mwiewior:
@codex pls review

mwiewior:
All three findings are fixed in `6f6eace`, each reproduced with a failing test first. The commit is on `chore/bump-to-header-parity-revs` (#46), which is stacked on this branch — see the placement note at the end.

**P1 — a directory holding only one side was skipped.** Confirmed: `discover_pairs` treated it as "not a run" and moved on, so a `--results-dir` gate could report PASS having quietly left out the contig whose annotation did not finish. A directory with *neither* side is still skipped (`_shared/` holds inputs, not outputs), but one side alone is now an error naming the directory and the missing pattern.

**P1 — a header mismatch could not fail the gate.** Fixed, and I want to be explicit about the shape, because gating on the whole header would have made the gate useless here rather than safer.

Concordance now requires the records **and the declarations**: `##INFO`, `##FORMAT`, `##FILTER`, `##ALT`, `##contig`, `##reference`, `##fileformat`, the `#CHROM` line, and any key the comparator does not recognise — unknown keys gate deliberately, so a header it does not understand fails loudly rather than passing silently. That is exactly your justification: those lines define how the record bytes are to be read, so matching bodies do not make two files interchangeable if an `##INFO` definition disagrees.

Lines describing how the *input* was produced — `##bcftools_*`, `##source`, `##fileDate` — are reported but not gated. In this corpus `##bcftools_normCommand` differs because the reference VEP output was made from a differently normalized input; nothing about interpreting a record depends on it, and gating on it would fail all 22 contigs for a harness artifact. The report now names which of the two a header difference is, and a new `DECL` column shows the gating verdict beside the body.

**P2 — strict mode normalized line endings.** Confirmed: `vcfio.open_text` decodes universal newlines, so a CRLF file and an LF file digested identically — against the whole claim of a mode that compares record bytes as written. Digests come from bytes now, via a new `vcfio.open_binary`. Canonical mode keeps ignoring line-ending style, which is the kind of cosmetic difference it exists to normalize, and there is a test for each half.

Verified on real data: chr21 still passes strict with its body digest unchanged (`a42bbf6468779da3c6ad0102c5bfd8e2`), now reporting `DECL ok` alongside the known `##bcftools_normCommand` difference. 1,116 Python tests, ruff clean.

**Placement.** The fixes and the comparator's documentation both sit in #46 because it is stacked here and already touches this file. If you would rather this PR merge with its own review closed, say so and I will cherry-pick both commits onto `e2e/md5-concordance` and rebase #46 on top — alternatively, close this one and let #46 carry the comparator, since it contains `2ad69c2` already.

PR #46, opened 2026-08-23
feat: byte-identical VCF output — preserve the input's INFO and FORMAT key order

mwiewior:
Closes the last gap between vepyr's VCF output and Ensembl VEP's.

> **Depends on** biodatageeks/datafusion-bio-formats#243 and biodatageeks/datafusion-bio-functions#216, pinned here by rev. Merge those and cut engine releases before this lands on a tag.

## Result

`md5_concordance.py --mode strict` now **passes on all 22 autosomes** of HG002 against Ensembl VEP 116.0 — **4,096,123 records**, byte-identical bodies.

| | chr21 | chr22 | all 22 |
|---|---|---|---|
| records | 55,812 | 50,861 | 4,096,123 |
| strict body | ✅ match | ✅ match | ✅ 22/22 |

Before this change strict mode reported three classes on every record:

```
55,812  FORMAT KEYS (-['PS'] +[])
55,812  SAMPLE1 keys
55,714  INFO order
```

## Why they could not be fixed in the writer

A record's INFO key order and its FORMAT key list are **per record**, and the engine reads a VCF into typed columns where every record carries every key the header declares, in schema order.

- **No file-level key order exists.** chr6 carries both `GT:PS:DP:ADALL:AD:GQ` and `GT:AD:PS` in one file.
- **`PS` cannot be recovered.** `Type=Integer` means a source `.` parses to null — identical to the key being absent. The writer inferred the key list from which values were non-missing and dropped it.

Ensembl VEP avoids the whole problem by never re-rendering: it copies the input line and appends to INFO (`OutputFactory/VCF.pm`).

## The change

The engine now carries two dictionary-encoded key-order columns from reader to writer (values are not carried — they already round-trip). `annotate()` gains `preserve_record_layout`, **on by default**, because byte agreement is the point of the VCF path. Passing `False` restores the previous output — verified byte-identical to the pre-change build's record body.

`AnnotateVcfConfig` is `#[non_exhaustive]` upstream now, so the config is built by assignment rather than a struct literal. That is what stops the next engine field addition from breaking this crate's build, as the last two did.

## Also here

The comparator excluded VEP's `##VEP*` provenance from the header digest but not vepyr's own `##datafusion-bio-function-vep*`, so identical output still reported a header difference. Both sides' provenance is excluded now, with tests that a real header difference still surfaces. chr21's header report goes from "234 vs 236 lines" with four differing lines to "234 vs 234" with one — the `##bcftools_normCommand` the reference run's different normalization left behind, which is a harness artifact and evidence the header passthrough works.

## Verification

- 1,110 Python tests (2 skipped), including the full golden suite with the flag defaulting on; clippy and ruff clean.
- 22/22 chromosomes in strict mode, run one at a time so the gate never needed more than one extra output on a disk already at 96%.
- Serial and `workers=4` produce the same body digest on chr21 (`a42bbf6468779da3c6ad0102c5bfd8e2`), so the sharded path does not diverge.

## Residual

QUAL is byte-correct for this corpus but not in general: it renders at f32 source precision, so an input written `50.0` returns as `50`. Every QUAL in HG002 is integral. Arbitrary-input parity still depends on datafusion-bio-formats#241.

https://claude.ai/code/session_01NifnHcLFdXTnLepZtnAxsB

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1150,8 +1150,8 @@ dependencies = [
 
 [[package]]
 name = "datafusion-bio-format-core"
-version = "1.10.0"
-source = "git+https://github.com/biodatageeks/datafusion-bio-formats.git?rev=7c866e07540aa8ae12d46e4878487911f659e8b9#7c866e07540aa8ae12d46e4878487911f659e8b9"
+version = "1.11.0"
+source = "git+https://github.com/biodatageeks/datafusion-bio-formats.git?rev=80f3a6c043899f8ded44eaab204b42735f805272#80f3a6c043899f8ded44eaab204b42735f805272"
 dependencies = [
  "async-compression",
  "bytes",
@@ -1173,8 +1173,8 @@ dependencies = [
 
 [[package]]
 name = "datafusion-bio-format-ensembl-cache"
-version = "1.10.0"
-source = "git+https://github.com/biodatageeks/datafusion-bio-formats.git?rev=7c866e07540aa8ae12d46e4878487911f659e8b9#7c866e07540aa8ae12d46e4878487911f659e8b9"
+version = "1.11.0"
+source = "git+https://github.com/biodatageeks/datafusion-bio-formats.git?rev=80f3a6c043899f8ded44eaab204b42735f805272#80f3a6c043899f8ded44eaab204b42735f805272"
 dependencies = [
  "async-trait",
  "datafusion",
@@ -1191,8 +1191,8 @@ dependencies = [
 
 [[package]]
 name = "datafusion-bio-format-vcf"
-version = "1.10.0"
-source = "git+https://github.com/biodatageeks/datafusion-bio-formats.git?rev=7c866e07540aa8ae12d46e4878487911f659e8b9#7c866e07540aa8ae12d46e4878487911f659e8b9"
+version = "1.11.0"
+source = "git+https://github.com/biodatageeks/datafusion-bio-formats.git?rev=80f3a6c043899f8ded44eaab204b42735f805272#80f3a6c043899f8ded44eaab204b42735f805272"
 dependencies = [
  "async-compression",
  "async-stream",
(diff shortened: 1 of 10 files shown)


mwiewior:
## Worker-scaling benchmark: no measurable cost

WGS HG002, merged cache, `everything` + `hgvs`, plain output, **4,096,123 records verified on every run**. M3 Max, single volume, run one worker count at a time so the ~59 GB peak (worker shards + assembled file) fits.

### Does the record-layout carry cost anything?

Three arms at 8 workers, best-of-n wall (the robust estimator under contention on this host):

| arm | wall best (n) | all samples | CPU s (mean) | peak RSS |
|---|---|---|---|---|
| carry **on** (shipping) | **70.9 s** (3) | 70.9, 73.8, 99.0 | 500.4 | 8.88 GiB |
| carry **off** (ablation, same binary) | 72.3 s (3) | 72.3, 72.4, 105.7 | 499.4 | 8.51 GiB |
| pre-change build (`c70860a` + bio-formats `9179628`) | 72.1 s (1) | 72.1 | 495.8 | 8.94 GiB |

All three sit within 2% of each other on wall and within 1% on total CPU. At 4 workers the same holds: 93.6 s on vs 97.3 s for the pre-change build, CPU 434.9 vs 432.2 (+0.6%).

**Total CPU-seconds is the metric to read here** — it is immune to the ambient load this host runs at, and it says the carry is free. The output grows by **5.0 bytes per record** (+0.07%), which is exactly the restored `PS` key and its value.

### Scaling curve, this branch

| workers | wall best (n=2, n=3 at w8) | speedup |
|---|---|---|
| 1 | 275.3 s | 1.00x |
| 2 | 145.8 s | 1.89x |
| 4 | 93.6 s | 2.94x |
| 8 | 70.9 s | 3.88x |

### Against the numbers in #44

#44's committed `summary.tsv` was measured on the Linux benchmark server (`/home/tgambin/...`), so the absolute seconds are not comparable to a Mac run — only the shape is:

| workers | #44 (server) | speedup | this branch (M3 Max) | speedup |
|---|---|---|---|---|
| 1 | 493.6 s | 1.00x | 275.3 s | 1.00x |
| 2 | 367.1 s | 1.34x | 145.8 s | 1.89x |
| 4 | 269.3 s | 1.83x | 93.6 s | 2.94x |
| 8 | 221.8 s | 2.23x | 70.9 s | 3.88x |

The curve is steeper here, which is expected: #44 predates the contig-prefetch work (#210, merged) that this branch builds on, and the hardware differs.

### Two measurement notes

- **The first WGS run of a session is an outlier.** Both 99.0 s (on) and 105.7 s (off) were the first two runs after an unrelated hour of I/O; every later sample at 8 workers landed at 70.9–73.8 s. Reading either of those n=1 would have produced a confident, wrong answer in both directions.
- **The 40 GiB guard is sized for the published SSD/HDD split.** On one volume a plain w>1 run peaks near 59 GB and APFS reclaims deleted outputs lazily, so six runs of the first sweep were correctly refused. The driver now waits for real free space before each run.

Raw metrics: `~/workspace/data_vepyr/archive/116/vepyr_{layout_on,layout_off,before_on,rep_on,repoff_off,curve2_on}/`.

PR #48, opened 2026-08-24
test(e2e): plugin comparison profiles + representation-only field equality

mwiewior:
Rebase and port of #40 onto current `master`. #40 branched from `8e3b9b1`; `master` has since moved to `7fcaf7a`, and `aa4fd01` split the monolithic `run_annotation_fast.py` into the `e2e-testing/scripts/comparison/` package, so all five of its files conflicted. This is the port, not a replay.

## What happened to each change in #40

| #40 change | Here |
|---|---|
| Pair CSQ entries by `Feature` instead of zipping two independently sorted lists | **Dropped — already on master in a stronger form.** `_pair_entry_groups()` pairs on `(ALLELE_NUM, Allele, Feature)`, removes exact-duplicate payloads first, and reports unmatched tails as one-sided entries instead of letting a transcript-set divergence cascade. |
| `values_equivalent()` — float formatting + `.` vs `""` | **Ported, tightened.** See below. |
| `merged_116_base` / `merged_116_5plugins` profiles | **Ported** as `merged_plugins_base` / `merged_plugins`, with derived paths. |
| `[patch]` override pointing `datafusion-bio-function-vep` at `../datafusion-bio-functions` | **Not ported.** #40 flagged it as unmergeable itself; `master` already pins a released tag. |
| `run_annotation_fast_all.py` `PROFILE_SUFFIXES` entries | Obsolete — that module and dict no longer exist; suffixes live on the `Profile` dataclass. |

## Representation-only field equality

Byte equality remains the only thing counted as a plain match. A difference that is representation-only is now absorbed into the match rate and counted in its own `field_format_mismatch_counts`, exactly the way `&`-list order differences already were, with its own `FormatOnly` column in the per-field table. Two cases:

- **Decimal padding.** vepyr prints Rust's shortest round-trip form (`0`, `0.57985`); the plugin sources Ensembl VEP passes through print a fixed number of decimals (`0.00`, `0.579850`). Same number, different string.
- **Missing-value marker.** VEP emits the VCF `.` for a field it has no value for, where vepyr emits the empty string.

Without this a plugin comparison is unreadable — #40 measured CADD_RAW at ~90% and SpliceAI `DS_*` at ~50% purely on formatting.

**Deliberate change from #40:** equivalence is exact float equality after parsing, not the `rel_tol=1e-4 / abs_tol=1e-6` epsilon the original used. That `abs_tol` would have quietly called a real gnomAD AF of 6e-7 equal to zero. Every case #40 documented (`"0"`/`"0.00"`, `"0.57985"`/`"0.579850"`) is exactly equal once parsed, so no tolerance is needed to get them. If a genuine printf-rounding residual turns up, it should get an explicit documented rule rather than a blanket epsilon.

## The parity gate stays exactly as strict

`verify_parity_gate.py` demands byte-identity, and that is unchanged:

- `field_format_mismatch_counts` is validated against the profile's field set and summed into a new `field_format_mismatch_total`. Any non-zero total still fails the gate — a released build must be byte-identical to Ensembl VEP.
- The equality-bucket invariant becomes `mismatches + order_only + format_only == strict_unequal`, so the new absorption path cannot hide a difference from the strict buckets.
- The counter is **required**, not defaulted to `{}`, so a report predating it cannot be read as zero.
- `expected_csq_fields()` now refuses a plugin profile outright with a clear message, rather than gating a field set it does not describe.

## Plugin profiles

- `merged_plugins` attaches the plugin cache — derived as `$DATA/cache/plugin_cache_{release}`, overridable with the new `--plugin-cache` — so ClinVar/SpliceAI/CADD/AlphaMissense/dbNSFP CSQ fields are comparable.
- `merged_plugins_base` reads the same five-plugin reference with nothing attached; the comparison then restricts to shared fields, isolating whether a core-field difference comes from the plugin machinery or predates it.

Both derive their paths like every other profile instead of hardcoding one machine's home directory, and a missing plugin cache fails in milliseconds with the availability matrix, same as a missing Parquet cache.

## Test plan

- [x] `pytest tests/` — 1130 passed, 2 skipped.
- [x] `ruff check` / `ruff format` clean; pre-commit hooks pass.
- [x] New unit coverage: representation-only equivalence (padding, exponent form, `&`-lists, `.`/`""`), the negatives it must **not** absorb (`0` vs `0.000001`, differing token counts, non-numeric), order-only precedence, plugin-profile resolution and its missing-cache failure, and four gate behaviours including that a format-only difference still fails.
- [x] **Re-measured, full chr1–22.** The open question above is now closed with this comparator, not #40's. Results in the section below.

### `chore(deps)`: pin bio-functions v0.19.0

`datafusion-bio-function-vep` moves off the #217 development head `1345fd5` to `01fa21f`, the commit named by the released **v0.19.0** tag. Both bio-formats dependencies stay on `80f3a6c` — the revision v0.19.0 itself pins; a mismatch would compile the formats crates twice and break type identity across the FFI boundary.

### `fix(e2e)`: only require plugin sources when a chromosome must be built

`run_all_plugin_comparisons.sh` probed the source server and required `bgzip`/`tabix` unconditionally, before the loop and under `set -e`. Those exist solely to slice sources for `build_transient_caches()`. With a complete plugin cache every iteration takes the `CACHE_REUSE` path and never opens a source — so the probe failed runs that needed nothing it was checking for. That is now the normal case, the full chr1–22 five-plugin cache having been built.

The slicing tools and the ClinVar `HEAD` probe are now gated on a `needs_sources` pass over `cache_complete()`, and the 1–22 validation is hoisted above it so a typo reports itself as an invalid chromosome rather than as a missing shard or a source-server error.

## Full-autosome parity results

Run from this branch against the Ensembl VEP 116 five-plugin references, engine `01fa21f`, reusing the complete plugin cache (no rebuilds):

| | |
|---|---|
| chromosomes | 22 (chr1–chr22) |
| VCF records | 4,096,123 |
| CSQ entries | 69,299,753 |
| strict body MD5 | **22/22 `ok`**, 22/22 exit 0 |
| fields at 100% | **124/124 on every chromosome** |

Zero on all 22, in every category: variants present on only one side, CSQ count / set / order mismatches, value mismatches, format-only mismatches, field-order mismatches, and mismatch-ledger rows.

Headers differ as expected (run provenance, absolute source paths, Ensembl's custom INFO declarations); record bodies are byte-identical.

**Scope:** autosomes only. chrX, chrY and chrMT are not covered by these caches or references.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1149,6 +1149,29 @@ dependencies = [
 ]
 
 [[package]]
+name = "datafusion-bio-format-bed"
+version = "1.11.0"
+source = "git+https://github.com/biodatageeks/datafusion-bio-formats.git?rev=80f3a6c043899f8ded44eaab204b42735f805272#80f3a6c043899f8ded44eaab204b42735f805272"
+dependencies = [
+ "async-compression",
+ "async-stream",
+ "async-trait",
+ "bytes",
+ "datafusion",
+ "datafusion-bio-format-core",
+ "datafusion-execution",
+ "env_logger",
+ "futures",
+ "futures-util",
+ "log",
+ "noodles-bed",
+ "noodles-bgzf 0.49.0",
+ "opendal",
+ "tokio",
+ "tokio-util",
+]
+
+[[package]]
 name = "datafusion-bio-format-core"
 version = "1.11.0"
 source = "git+https://github.com/biodatageeks/datafusion-bio-formats.git?rev=80f3a6c043899f8ded44eaab204b42735f805272#80f3a6c043899f8ded44eaab204b42735f805272"
@@ -1223,7 +1246,7 @@ dependencies = [
 [[package]]
 name = "datafusion-bio-function-vep"
 version = "0.16.0"
-source = "git+https://github.com/biodatageeks/datafusion-bio-functions.git?rev=e937e49ace4b4ba997a53be05da7bd7d6db0d399#e937e49ace4b4ba997a53be05da7bd7d6db0d399"
+source = "git+https://github.com/biodatageeks/datafusion-bio-functions.git?rev=86fd72d4b716fd3cdd2d85520ab81dea325507be#86fd72d4b716fd3cdd2d85520ab81dea325507be"
 dependencies = [
  "ahash",
  "async-trait",
@@ -1231,6 +1254,7 @@ dependencies = [
  "chrono",
  "coitrees",
  "datafusion",
+ "datafusion-bio-format-bed",
  "datafusion-bio-format-core",
  "datafusion-bio-format-ensembl-cache",
  "datafusion-bio-format-vcf",
@@ -3162,6 +3186,20 @@ dependencies = [
  "noodles-vcf",
  "pin-project-lite",
  "tokio",
+]
+
+[[package]]
+name = "noodles-bed"
+version = "0.36.0"
+source = "git+https://github.com/biodatageeks/noodles.git?rev=42a3c0165617b7b8f9f5c5905287968eef92e31e#42a3c0165617b7b8f9f5c5905287968eef92e31e"
+dependencies = [
+ "bstr",
+ "lexical-core",
+ "memchr",
+ "noodles-bgzf 0.49.0",
+ "noodles-core 0.20.0",
+ "noodles-csi",
+ "noodles-tabix",
 ]
 
 [[package]]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1245,8 +1245,8 @@ dependencies = [
 
 [[package]]
 name = "datafusion-bio-function-vep"
-version = "0.16.0"
-source = "git+https://github.com/biodatageeks/datafusion-bio-functions.git?rev=1345fd5ca5b7dc097825813b05856ad8c3bca578#1345fd5ca5b7dc097825813b05856ad8c3bca578"
+version = "0.17.0"
+source = "git+https://github.com/biodatageeks/datafusion-bio-functions.git?rev=01fa21f8533de17aeffa5ee9042381c4171ef02b#01fa21f8533de17aeffa5ee9042381c4171ef02b"
 dependencies = [
  "ahash",
  "async-trait",
@@ -1266,7 +1266,10 @@ dependencies = [
  "log",
  "nohash-hasher",
  "noodles-core 0.16.0",
+ "noodles-core 0.20.0",
+ "noodles-csi",
  "noodles-fasta",
+ "noodles-tabix",
  "parquet",
  "rapidhash",
  "serde",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1246,7 +1246,7 @@ dependencies = [
 [[package]]
 name = "datafusion-bio-function-vep"
 version = "0.16.0"
-source = "git+https://github.com/biodatageeks/datafusion-bio-functions.git?rev=86fd72d4b716fd3cdd2d85520ab81dea325507be#86fd72d4b716fd3cdd2d85520ab81dea325507be"
+source = "git+https://github.com/biodatageeks/datafusion-bio-functions.git?rev=1345fd5ca5b7dc097825813b05856ad8c3bca578#1345fd5ca5b7dc097825813b05856ad8c3bca578"
 dependencies = [
  "ahash",
  "async-trait",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,17 +42,19 @@ log = "0.4"
 env_logger = { version = "0.11", default-features = false }
 serde_json = "1"
 
-# Released bio-formats v1.11.0 and bio-functions v0.18.0, pinned by the commit
+# Released bio-formats v1.11.0 and bio-functions v0.19.0, pinned by the commit
 # each tag names. v1.11.0 carries each record's INFO and FORMAT key order through
-# a read/write round trip; v0.18.0 turns that on for annotation output and stops
-# stripping field metadata from the annotation schema, without which the carry
-# reaches the writer unrecognised. Together with the merged header and QUAL work
-# they make vepyr's VCF output byte-identical to Ensembl VEP 116.0 on all 22
-# autosomes of HG002.
+# a read/write round trip; v0.18.0 turned that on for annotation output and
+# stopped stripping field metadata from the annotation schema, without which the
+# carry reaches the writer unrecognised. v0.19.0 adds the plugin cache-builder
+# work: source-level tabix selection for indexed delimited sources, multi-part
+# manifests, and registration of the tabix staging slice as headerless. Together
+# with the merged header and QUAL work they make vepyr's VCF output
+# byte-identical to Ensembl VEP 116.0 on all 22 autosomes of HG002.
 # The two MUST move together: bio-function-vep pins its own bio-formats rev, and
 # two different revs are two different sources to cargo, so a mismatch compiles
 # the formats crates twice and `CacheSourceType` / the VCF writer types then fail
 # to cross into `CacheBuilder` (E0308).
-datafusion-bio-function-vep = { git = "https://github.com/biodatageeks/datafusion-bio-functions.git", rev = "1345fd5ca5b7dc097825813b05856ad8c3bca578", features = ["cache-builder"] }
+datafusion-bio-function-vep = { git = "https://github.com/biodatageeks/datafusion-bio-functions.git", rev = "01fa21f8533de17aeffa5ee9042381c4171ef02b", features = ["cache-builder"] }
 datafusion-bio-format-ensembl-cache = { git = "https://github.com/biodatageeks/datafusion-bio-formats.git", rev = "80f3a6c043899f8ded44eaab204b42735f805272" }
 datafusion-bio-format-vcf = { git = "https://github.com/biodatageeks/datafusion-bio-formats.git", rev = "80f3a6c043899f8ded44eaab204b42735f805272" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -53,8 +53,6 @@ serde_json = "1"
 # two different revs are two different sources to cargo, so a mismatch compiles
 # the formats crates twice and `CacheSourceType` / the VCF writer types then fail
 # to cross into `CacheBuilder` (E0308).
-datafusion-bio-function-vep = { git = "https://github.com/biodatageeks/datafusion-bio-functions.git", rev = "e937e49ace4b4ba997a53be05da7bd7d6db0d399", features = ["cache-builder"] }
+datafusion-bio-function-vep = { git = "https://github.com/biodatageeks/datafusion-bio-functions.git", rev = "86fd72d4b716fd3cdd2d85520ab81dea325507be", features = ["cache-builder"] }
 datafusion-bio-format-ensembl-cache = { git = "https://github.com/biodatageeks/datafusion-bio-formats.git", rev = "80f3a6c043899f8ded44eaab204b42735f805272" }
 datafusion-bio-format-vcf = { git = "https://github.com/biodatageeks/datafusion-bio-formats.git", rev = "80f3a6c043899f8ded44eaab204b42735f805272" }
-
-
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -53,6 +53,6 @@ serde_json = "1"
 # two different revs are two different sources to cargo, so a mismatch compiles
 # the formats crates twice and `CacheSourceType` / the VCF writer types then fail
 # to cross into `CacheBuilder` (E0308).
-datafusion-bio-function-vep = { git = "https://github.com/biodatageeks/datafusion-bio-functions.git", rev = "86fd72d4b716fd3cdd2d85520ab81dea325507be", features = ["cache-builder"] }
+datafusion-bio-function-vep = { git = "https://github.com/biodatageeks/datafusion-bio-functions.git", rev = "1345fd5ca5b7dc097825813b05856ad8c3bca578", features = ["cache-builder"] }
 datafusion-bio-format-ensembl-cache = { git = "https://github.com/biodatageeks/datafusion-bio-formats.git", rev = "80f3a6c043899f8ded44eaab204b42735f805272" }
 datafusion-bio-format-vcf = { git = "https://github.com/biodatageeks/datafusion-bio-formats.git", rev = "80f3a6c043899f8ded44eaab204b42735f805272" }
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -101,8 +101,19 @@ Two rules are in use upstream:
 
 | rule | what Ensembl does | plugins |
 |---|---|---|
-| `minimised` | calls `get_matched_variant_alleles()`, which runs `trim_sequences()` over **both** the variant and the data row — trimming shared prefix *and* suffix, in both orders — then compares `(ref, alt, pos)` | CADD, AlphaMissense |
-| `exact` | compares `(start, ref, alt)` verbatim | SpliceAI, dbNSFP, and core `--custom ...,exact` (ClinVar) |
+| `minimised` | calls `get_matched_variant_alleles()`, which runs `trim_sequences()` over **both** the variant and the data row — trimming shared prefix *and* suffix, in both orders — then compares `(ref, alt, pos)` | CADD, AlphaMissense, ClinVar |
+| `exact` | compares `(start, ref, alt)` verbatim | SpliceAI, dbNSFP |
+
+**ClinVar is `minimised`, despite being loaded with `--custom ...,exact`.** The
+`exact` there names the *overlap mode*, not the allele rule — core still
+minimises the alleles first, then requires exact overlap. The golden VEP 116
+reference settles it: `chr1:65364614 GT>TT` is annotated `ClinVar=1258041`,
+whose source record is the 1 bp SNV `G>T` at the same position. Reaching it
+from `GT/TT` needs the shared trailing `T` trimmed, so this is full
+`trim_sequences()` behaviour — a suffix trim, not just the narrower
+leading-anchor-base shift. A verbatim comparison would emit nothing, and a
+1 bp feature could not have exactly overlapped the 2 bp unminimised variant
+either.
 
 The distinction bites whenever a record is **not in minimal form**. `bcftools
 norm -m -both` splits a multi-allelic record without re-trimming its halves, so
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -85,8 +85,72 @@ source and map it to CSQ fields.
   plugins (the value is emitted on every transcript line).
 - **`[[value_columns]]`** *(1+)* — `column`, `csq_field` (output field name),
   `type` (`Utf8` / `Float32` / `Int32`). **Declaration order = CSQ output order.**
+- **`allele_match`** *(optional)* — `"exact"` (default) or `"minimised"`. Which
+  one is correct is decided by the plugin's own Ensembl implementation, not by
+  preference; see [Allele matching](#allele-matching-exact-vs-minimised).
 
 There is **no `[tier]` block** — tiering is inherited from the variation cache.
+
+## Allele matching: `exact` vs `minimised`
+
+Ensembl's plugins do **not** agree on how a variant is compared to a row of
+their data file, so neither can vepyr. Getting this wrong is silent: the wrong
+setting either drops annotations Ensembl reports, or invents ones it does not.
+
+Two rules are in use upstream:
+
+| rule | what Ensembl does | plugins |
+|---|---|---|
+| `minimised` | calls `get_matched_variant_alleles()`, which runs `trim_sequences()` over **both** the variant and the data row — trimming shared prefix *and* suffix, in both orders — then compares `(ref, alt, pos)` | CADD, AlphaMissense |
+| `exact` | compares `(start, ref, alt)` verbatim | SpliceAI, dbNSFP, and core `--custom ...,exact` (ClinVar) |
+
+The distinction bites whenever a record is **not in minimal form**. `bcftools
+norm -m -both` splits a multi-allelic record without re-trimming its halves, so
+this is routine rather than exotic:
+
+```text
+chr21:13973877  REF=TTGTGTGTGTGTG  ALT=GTGTGTGTGTGTG   # really just T>G
+chr21:26062230  REF=AAC            ALT=ACAC            # really just an inserted C
+```
+
+CADD keys the first as `T/G` in its per-base file. Under `minimised` Ensembl
+reduces the variant and matches; under `exact` it does not — and SpliceAI, whose
+rows at that position are plain SNVs, reports nothing. Both are correct *for
+their own plugin*.
+
+### Trim order is load-bearing
+
+`trim_sequences()` can trim prefix-first or suffix-first, and for a non-minimal
+indel **the two orders land on different coordinates — and therefore different
+variants**. Ensembl's VCF parser builds the `VariationFeature` prefix-first
+(left-first), so that is the only order that reproduces its output:
+
+| VCF record | left-first (Ensembl) | right-first |
+|---|---|---|
+| `CGTGTGT/CGTGT` | `GT/-` at 13836153 — no row there, so empty, as VEP reports | `GT/-` at 13836149 — a **different variant's** score |
+| `AAC/ACAC` | `-/C` at 26062231 — the row VEP reports | same |
+
+Getting this backwards is silent and produces confident wrong answers: on chr21
+a right-first reduction invented 5,288 CADD scores VEP does not emit, and
+regressed ClinVar and SpliceAI from clean.
+
+!!! warning "Never set `minimised` to gain hits"
+    Reducing alleles for an `exact` plugin produces annotations Ensembl does not
+    emit — SpliceAI gained 68 spurious hits on chr21 that way. The setting is a
+    statement about what upstream does, not a tuning knob.
+
+To decide the setting for a new plugin, read its `.pm`: if `run()` calls
+`get_matched_variant_alleles`, use `minimised`; if it compares `$vf->{start}`
+and the allele strings with `==`/`eq`, use `exact`.
+
+!!! note "Ensembl's fetch window"
+    `get_matched_variant_alleles` is not the whole story upstream. CADD.pm first
+    fetches `[VF.start - 2, VF.end]` from the tabix file, so a row whose *file*
+    position falls outside that window is never even considered — which is why
+    VEP reports nothing for `CGTGTGT/CGTGT` even though a `GT/-` row exists
+    nearby. Left-first reduction happens to land outside that row's reach too,
+    so vepyr agrees without emulating the window; if a future mismatch traces
+    back here, this is the mechanism to check.
 
 ### Example: AlphaMissense (per-transcript)
 
--- a/docs/testing-vep.md
+++ b/docs/testing-vep.md
@@ -398,9 +398,17 @@ uv run python run_comparison.py --release 115
 # A specific pick-mode profile
 uv run python run_comparison.py --release 115 --profile merged_pick_allele_gene
 
-# Plugin fields, against a VEP reference built with the same five plugins
-uv run python run_comparison.py --release 116 --profile merged_plugins
+# Plugin fields, against a VEP reference built with the same five plugins.
+# Plugin references are written one file per contig, so name the contig.
+uv run python run_comparison.py --release 116 --profile merged_plugins --chroms 22
 ```
+
+!!! note "Plugin profiles need a contig"
+    `generate_vep_plugin_references.sh` writes one reference per contig under
+    `output/{release}/plugins/`, so unlike the whole-genome profiles there is no
+    single file to slice. Pass a single `--chroms` value, or `--vep` explicitly.
+    Modes that read no reference at all, such as `--skip-annotate` aggregation
+    of stored reports, do not need either.
 
 !!! note "Plugin profiles"
     `merged_plugins` attaches the plugin cache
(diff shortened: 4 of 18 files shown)


mwiewior:
Pushed 3e7f805 closing three findings from a review of the first commit.

**`_tokens_equivalent` used `float()`, which absorbs genuine differences.** The docstring claimed it never calls two different numbers equal; `float()` breaks that in five ways, all verified against the real function:

| | |
|---|---|
| `'12345678901234567'` vs `'12345678901234568'` | same double past 2^53 |
| `'01'` vs `'1'` | zero-padded identifier |
| `' 1'` vs `'1'` | whitespace |
| `'1_0'` vs `'10'` | Python's digit separator |
| `'1e400'` vs `'1e999'` | both become `inf` |

A plugin field carrying a long numeric ID or a zero-padded code would have had two distinct values written off as formatting. Now exact `Decimal` equality over a canonical numeric literal — no whitespace, no separators, no `Infinity`/`NaN`, no leading zero on the integer part. All five now compare unequal; `"0"`/`"0.00"`, `"0.57985"`/`"0.579850"`, `"1e-05"`/`"0.00001"` still absorb.

**The all-match banner printed while byte differences existed.** `ALL n shared CSQ fields match at 100%!` was gated on `field_mismatches` alone, so a run with order-only or format-only differences — one the parity gate rejects — printed the exact string an operator greps for to conclude parity. Those fields now get their own block stating it is not byte parity; the banner prints only on real parity.

**`""`/`"."` was symmetric.** VEP writes the VCF missing marker where it has no value and vepyr writes the empty string, so only that direction is a representation difference — a `.` coming *from* vepyr is an output defect. Made directional.

1138 tests pass (was 1130); ruff clean.

Four lower-severity findings not addressed, for the record: format-only differences emit no ledger row (so a gate failure gives an empty ledger to debug from), `field_format` is aggregated in `report.py` but never rendered in the multi-contig summary (`field_order` has the same pre-existing gap), `--plugin-cache` is silently ignored for non-plugin profiles, and the declared `plugins` tuple is not validated against the cache's actual contents.

mwiewior:
@codex pls review

mwiewior:
@claude pls review

mwiewior:
Pushed `d6ac388`, `6937eed`, `b057c3b` addressing all five Codex findings. Each was verified against the code before implementing — none were dismissed.

**P1 — pre-existing shards destroyed (`run_all_plugin_comparisons.sh`).** Confirmed and fixed. `cache_complete()` demands all five shards, so a chromosome missing one rebuilt all five with `overwrite=True`, and cleanup unlinked all five — losing four pre-existing shards on a *successful* run, contradicting the script header's promise to "preserve the cache shards that existed at startup". Now records the shard paths present before the build and keeps exactly those. Verified with a simulated partial cache (4 of 5 plugins cached for chr5): 4 survive, only the genuinely transient shard is removed. Before the fix all 5 were deleted.

**P2 — early exit leaked transient state.** Confirmed and fixed. Worth noting this was worse than disk exhaustion: `cache_complete()` tests file existence alone, so orphaned shards left behind by a failed run would be *reused* on a later run against a manifest the trap had already reverted. The trap now drops the in-flight chromosome's artifacts before restoring manifests.

**P2 — data root not propagated.** Confirmed and fixed. The runner read `VEPYR_DATA_ROOT` while `comparison/profiles.py` reads `DATA_VEPYR_DIR` — two different names, so overriding the runner's root moved only the two paths passed as explicit flags. Now accepts either name and exports `DATA_VEPYR_DIR` so both agree.

**P2 — benchmark VCF path.** Confirmed and fixed. I initially doubted this one, since the benchmark sits at the data root on the machine these runs were done on — but `docs/testing-vep.md` documents `$DATA/input/` as the layout, so that machine is the legacy exception and the finding is right. Now mirrors the `_resolve_with_legacy_fallback` behaviour the Python profiles already use, and names the documented path in the failure when neither exists.

**P2 — format-only differences invisible in aggregates.** Confirmed and fixed. `aggregate_mismatches()` collected the counts but `generate_markdown()` derived the headline solely from `field_mismatch_counts`. Now reported as its own explicitly non-parity line, with the existing headline qualified as *value* mismatches. `n_perfect` keeps its meaning — format-only differences are absorbed into the match rates by design — so the gate's semantics are unchanged. Two regression tests added.

Full suite: 1140 passed, 2 skipped. `ruff check` clean; both shell scripts pass `bash -n`.

Not addressed, deliberately: the `/Users/mwiewior/...` defaults in the runner (local dev harness, env-overridable, and now reachable via either variable name) and the pre-existing `ruff format` drift in `performance-tests/vep/scripts/plot_vep_fork_scaling.py`, which came from #44 on master and is outside this PR.

@codex pls review
@claude pls review

mwiewior:
Round two: pushed `822280c` and `9f85920` for the three new findings against `b057c3b`. All three verified against the code first.

**P2 — builders not reaped before cleanup (`run_all_plugin_comparisons.sh:82`).** Confirmed, and a genuine gap in my own trap fix from `d6ac388` — thank you, this is the good kind of finding. The handler cleaned up and then *returned*, so bash resumed the interrupted `wait` and the three background builders were free to recreate a shard after cleanup, while `CURRENT_TRANSIENT=0` made the EXIT trap skip a second pass.

I reproduced it before fixing, with a harness mimicking the structure — a builder that writes its shard 3s in, interrupted at 1s:

- **before:** handler cleaned, script resumed to `COMPLETED NORMALLY` with `rc=0`, and the shard reappeared at t=3 — exactly the orphaned state the trap was meant to prevent
- **after:** builder reaped, shard gone, `rc=143`, nothing reappears

INT/TERM now reap the tracked child PIDs and `exit 128+signal`; the EXIT trap then does the single cleanup with nothing still running.

**P2 — concurrent writers to the normalized VCF (`generate_vep_plugin_references.sh:75`).** Confirmed: `VEP_REFERENCE_JOBS` chromosomes launch at once and on a fresh data root every one fails the `-f "$NORM"` check, then races on `bcftools norm -o` and `tabix -f` against the same paths. Took the atomic-publish option you offered: build under a unique name, then `mv` into place — index first, since `$NORM` appearing is what the existence check keys on.

**P2 — work directories outside `$DATA` (`build_vep_plugin_reference.sh:208`).** Confirmed. `WORK` comes from an advertised optional positional argument, so pointing it outside `$DATA` made `${IN#"$DATA/"}` a no-op yielding `/data//abs/host/path`, which VEP reports as a missing input rather than the configuration error it is. Every host path now goes through a `docker_path()` helper that translates or fails loudly. Verified both directions: a path under `$DATA` maps to `/data/...`, one outside is rejected with the reason.

Full suite: 1140 passed, 2 skipped. `ruff check` clean; all three shell scripts pass `bash -n`.

Still deliberately not addressed, both previously noted: the `/Users/mwiewior/...` defaults in the local dev runner, and the pre-existing `ruff format` drift in `performance-tests/vep/scripts/plot_vep_fork_scaling.py` (from #44 on master, outside this PR).

Also noted from Claude's review and left as a follow-up: `build_transient_caches()` still rebuilds all five plugins when only one is missing. That is an inefficiency, not a correctness bug — the P1 fix stops pre-existing shards being deleted, it does not skip the redundant rebuild.

@codex pls review
@claude pls review

mwiewior:
Round three: pushed `0b89fe0` and `fd116f3` for the three findings against `9f85920`. All three real, all reproduced before fixing.

**P2 — pre-existing shards still overwritten (`:197`).** You're right, and this is the more serious of the two shard findings: my `d6ac388` preservation recorded *paths*, but the rebuild had already overwritten the *contents*. What survived cleanup was the rebuilt file, left paired with the manifest `restore_manifests()` puts back — so a run could leave rewritten shards beside stale metadata, and the next path-only `cache_complete()` would reuse them.

Fixed at the source rather than the symptom: `build_transient_caches()` now builds only genuinely-absent plugins. The kept shard is then the original and still matches its manifest. This also removes the redundant re-slicing that Claude's review flagged separately as an inefficiency — one change, both problems.

Verified: with 4 of 5 plugins cached for a chromosome, only the missing one is built and all four originals still hold their original contents.

**P2 — only the outer builder shell was killed (`:84`).** Confirmed. Each PID names the `build_x ... &` subshell; the real work runs in nested subshells (tabix, bgzip, `uv run python`). Reproduced with a builder whose work runs in a nested `sh -c`:

- plain `kill` of the outer PID → **child and grandchild both still alive**
- `kill_tree()` → all three gone, no orphaned output afterwards

`kill_tree()` walks descendants depth-first via `pgrep -P`, so children die before their parent.

**P2 — order-only differences invisible (`report.py:510`).** Confirmed, and a fair "you fixed half of it": `b057c3b` threaded `field_format` into the headline but left `field_order`, though aggregation had been collecting it all along. Same treatment, same wording, same pair of tests (present when non-zero, absent otherwise).

Full suite: 1142 passed, 2 skipped. `ruff check` clean; all three shell scripts pass `bash -n`.

A note on the earlier signal work, since it bears on how much to trust that verification: my first attempt to test the trap used `kill -INT` on a backgrounded script and reported PASS — but bash ignores SIGINT for background jobs, so the handler never ran and the test proved nothing. The SIGTERM rerun exposed the real behaviour. Worth stating plainly given the round-2 claim rested on it.

Unchanged from before: the `/Users/mwiewior/...` defaults in the local dev runner, and the pre-existing `ruff format` drift in `performance-tests/vep/scripts/plot_vep_fork_scaling.py` (from #44 on master, outside this PR).

@codex pls review
@claude pls review

mwiewior:
### `allele_match` table lists ClinVar as `exact`, but the manifest ships `minimised`

Found while cross-checking this branch against the plugin caches published in #52.

`docs/plugins.md:105` puts ClinVar in the `exact` row:

> | `exact` | compares `(start, ref, alt)` verbatim | SpliceAI, dbNSFP, and core `--custom ...,exact` (ClinVar) |

But `clinvar.source.toml` sets `minimised` explicitly — on **every** branch of `vepyr-plugins` (`master`, `plugin-test-fixes`, `fix/dbnsfp-gzip-source`, `fix/indexed-bgzf-text-sources`, `fix/adding-a-plugin-sort-guidance`), and with an empirical justification attached:

```toml
# VEP's --custom VCF source minimises alleles before comparing (verified on
# chr1:65364614 GT>TT, which it matches to ClinVar's G>T row 1258041). The
# "exact" in --custom ...,exact is the overlap mode, not the allele rule.
allele_match      = "minimised"
```

The built caches follow the manifest, so the discrepancy reaches the published artifacts:

| plugin | `.source.toml` | built `manifest.json` | this PR's table |
|---|---|---|---|
| **clinvar** | **`minimised`** (explicit) | **`minimised`** | **`exact`** ❌ |
| spliceai | *(default)* | `exact` | `exact` ✅ |
| dbnsfp | *(default)* | `exact` | `exact` ✅ |
| cadd | `minimised` | `minimised` | `minimised` ✅ |
| alphamissense | `minimised` | `minimised` | `minimised` ✅ |

ClinVar is the only row that disagrees, and it is the only one where the manifest states the value explicitly rather than falling through to the `exact` default — so this is unlikely to be an accidental setting.

Which way it resolves matters, because the two sources make opposite claims about the same thing — whether `--custom ...,exact` constrains the *allele rule* or only the *overlap mode*:

- **If the table is right**, the four published ClinVar caches were built under the wrong rule and need a rebuild. By this PR's own warning ("Never set `minimised` to gain hits"), `minimised` on an `exact` plugin yields annotations Ensembl does not emit.
- **If the toml is right**, the table's ClinVar entry needs correcting before it becomes the reference people read — the same section notes this class of error is silent.

The toml carries a specific reproducer (chr1:65364614 `GT>TT` → ClinVar row 1258041) and this PR's harness has the golden references to settle it either way; I have not re-run that comparison, so I am reporting the inconsistency rather than a verdict.

Cross-ref: #52 documents these caches for download, and its ClinVar dataset card currently states `minimised`, matching the shipped manifest. If the table wins, that card needs updating alongside the rebuild.

---

mwiewior:
### Resolved against the golden reference: ClinVar is `minimised`

Checked the reproducer from `clinvar.source.toml` against the release-116 five-plugin VEP reference (`output/116/plugins/HG002_chr1_5plugins_vep116_caddfix.vcf.gz`). **The toml is right and the table row is wrong** — no rebuild is needed.

**The variant.** `chr1:65364614 GT>TT` is present in the reference.

**What real VEP 116 emits** — all 10 CSQ entries for the `TT` allele carry ClinVar:

```
ClinVar=1258041
ClinVar_CLNSIG=Benign
ClinVar_CLNREVSTAT=criteria_provided&_multiple_submitters&_no_conflicts
ClinVar_CLNDN=not_provided&Juvenile_onset_Parkinson_disease_19A
ClinVar_CLNVC=single_nucleotide_variant
ClinVar_CLNVI=ClinGen:CA893568
```

**Why that settles it.** The matched record is an SNV in the source VCF:

```
1  65364614  1258041  G  T  ...  CLNHGVS=NC_000001.11:g.65364614G>T
```

So VEP matched a 2 bp→2 bp variant (`GT/TT`) to a 1 bp→1 bp row (`G/T`). That requires trimming the shared trailing `T` — i.e. minimisation. Under a verbatim `(start, ref, alt)` comparison the probe key `(65364614, "GT/TT")` cannot reach `(65364614, "G/T")`, and VEP would emit nothing. It emits the annotation, so the allele rule is `minimised`.

Note this is a **suffix** trim, not the leading-anchor-base trim — so it is full `trim_sequences()` behaviour, not the narrower `vcf_to_vep_input_allele()` shift.

There is a second, independent signal in the same record. The ClinVar feature spans 65364614–65364614 while the unminimised VF spans 65364614–65364615. Had `exact` constrained coordinates verbatim, a 1 bp feature could not have matched a 2 bp VF either. The match is only consistent with **minimise the alleles first, then apply exact overlap** — which is exactly the toml's claim that "the `exact` in `--custom ...,exact` is the overlap mode, not the allele rule."

**Consequences**

- `docs/plugins.md:105` should move ClinVar out of the `exact` row. Suggested: keep `exact` as *"SpliceAI, dbNSFP"* and add ClinVar to `minimised` with a parenthetical that core `--custom ...,exact` names the overlap mode, not the allele rule — the distinction is the whole trap here.
- The published ClinVar cache is **correct as built**. It stores this row as `start=65364614, allele_string=G/T, id=1258041, clnsig=Benign, clnvc=single_nucleotide_variant`, which reproduces the golden output exactly. No rebuild, and the dataset card in #52 needs no change.

Method, so it can be re-run: `tabix` the reference at `chr1:65364614`, split the `CSQ=` field on `,`, and read fields 119–124 (`ClinVar` … `ClinVar_CLNVI`) of the 124-field CSQ header.

---

mwiewior:
Confirmed independently and fixed in `8ce05c8` — docs only, no cache or code change.

I re-ran the reproducer rather than taking the conclusion as given, and it holds on every step:

- `chr1:65364614 GT>TT` carries `ClinVar=1258041` on all **10** of its CSQ entries in the release-116 golden reference.
- The matched record really is a 1 bp SNV. From the built cache: `chrom=1, start=65364614, end=65364614, allele_string=G/T, id=1258041, clnsig=Benign, clnvc=single_nucleotide_variant, clnvi=ClinGen:CA893568`.
- So `("GT","TT")` reached `("G","T")`, which needs the shared trailing `T` trimmed — a **suffix** trim, i.e. full `trim_sequences()`, not the narrower leading-anchor-base shift.
- The coordinate signal agrees: a 1 bp feature (65364614–65364614) cannot exactly overlap the 2 bp unminimised VF (65364614–65364615).

One correction to the method as written: `chrom` is stored as `1`, not `chr1`, so a filter on `chr1` returns zero rows and looks like the row is absent. Worth knowing for anyone re-running it.

The manifest has always declared `allele_match = "minimised"`, so `docs/plugins.md:105` was the only thing claiming otherwise — the doc had read the `exact` in `--custom file,ClinVar,vcf,exact,...` as the allele rule when it names the overlap mode. ClinVar now sits in the `minimised` row, `exact` keeps SpliceAI and dbNSFP, and a short paragraph records the distinction and the evidence so the trap does not get re-introduced.

Agreed on the rest: the published cache is correct as built and the #52 dataset card needs no change.

mwiewior:
Round four: pushed `e9e453a` for both findings against `fd116f3`.

**P2 — descendants could outlive `terminate_builders()` (`:96`).** Confirmed, and I had to work to prove it — my first two attempts at a reproducer did not discriminate, because `kill_tree()` kills depth-first and the descendant usually died with its `sleep`. The case that does separate them is a builder whose *outer shell exits immediately*, leaving a detached worker: `pgrep -P` then has no parent link to follow, and `wait` reaps only the recorded child.

- previous implementation → **shard reappears after cleanup**
- this one → clean

Builders now launch under `set -m` so each leads its own process group; the group is signalled as a unit, then polled until genuinely empty with escalation to KILL. `kill_tree()` remains the fallback when group signalling is unavailable. Verified the mechanism directly first: a `set -m` background job does become its own process-group leader here, and `kill -- -PGID` reaches the nested tree.

**P2 — `CURRENT_PIDS` published too late (`:289`).** Confirmed by inspection, no race harness needed — it was assigned after the launch loop finished, so a signal arriving mid-loop saw the pre-loop value and reaped nothing. Each PID is now appended as its own launch happens.

While there: SpliceAI and CADD move from foreground to backgrounded-then-waited. They remain sequential, but are now tracked like the other three, and a signal during a long CADD build is handled immediately instead of after it completes. That was not in the finding; it is the same hole, and leaving it would have meant the fix only covered three of five builders.

Full suite: 1142 passed, 2 skipped. `ruff check` clean; all shell scripts pass `bash -n`.

Also in this push, from a separate reviewer thread: `8ce05c8` corrects `docs/plugins.md` to list ClinVar's allele rule as `minimised` rather than `exact`, verified against the golden reference. Docs only — the manifest already declared `minimised` and the published cache is correct as built.


@codex pls review
@claude pls review

mwiewior:
Pushed `c9bd75f`. Taking the hang finding seriously enough to test it properly, then hardening even though it did not reproduce.

**The hang does not reproduce.** You correctly flagged the analysis as static and asked for a live repro before trusting it, so I ran one. `terminate_builders()` called **directly in the main shell** (not backgrounded — my first harness got that wrong, which would have invalidated the result), against four builder shapes:

| builder | darwin | Linux (bash:5) |
|---|---|---|
| normal, dies on TERM | returns 0–1s | returns 0s |
| nested worker | returns 0s | returns 1s |
| ignores TERM (forces KILL) | returns 10s | returns 10s |
| detached worker, parent exits first | returns 0s | returns 0s |

I also checked the specific premise directly: after TERM, `kill -0 -- "-$PGID"` **fails** before any `wait` on both platforms — bash reaps background children asynchronously, so the leader's zombie does not linger to keep the group non-empty.

**Hardened regardless.** The ordering costs nothing, the failure mode you describe is unrecoverable (SIGKILL bypasses the EXIT trap, so *none* of the shard preservation or manifest restoration from this thread would run), and I cannot verify every bash/libc a contributor might have. So: reap the builder first, then poll the group for real survivors. Both loops now `break` after escalating to KILL, so neither can spin regardless of how a platform accounts for zombies — the unbounded `deadline=$((SECONDS + 5))` re-arm is gone.

Re-verified the invariant after the reorder, since I changed the ordering the earlier evidence rested on: a detached worker whose parent exits before the kill walk still cannot recreate a shard after cleanup — plain, TERM-ignoring, and detached cases all clean.

Thanks also for the confirmation on `set -m`, the early `CURRENT_PIDS` publication, and the SpliceAI/CADD tracking — and for independently reaching the same process-group remedy you'd suggested as the TOCTOU fix in the previous round.

Full suite: 1142 passed, 2 skipped. `ruff check` clean; all shell scripts pass `bash -n`.

@codex pls review
@claude pls review

mwiewior:
Round six: pushed `a6c1b10`, `f3c9c21`, `103440f`, `4908d8b` for all five findings (two from the `e9e453a` pass I had not yet answered, three from `c9bd75f`). All verified against the code first.

**P2 — `..` defeated the containment check (`build_vep_plugin_reference.sh:205`).** Confirmed: my round-two `docker_path()` used a lexical prefix test, so `$DATA/../scratch` passed it and became `/data/../scratch` → `/scratch` inside the container. My first fix attempt was still bypassable — `cd`-based resolution leaves a non-existent directory untouched, so the prefix kept matching. `..` and `.` are now collapsed lexically as well. Verified: escape rejected, outside-path rejected, and a `..` that stays inside `$DATA` still resolves and is accepted.

**P2 — reused references not bound to plugin bytes (`generate_vep_plugin_references.sh:31`).** Confirmed, and the gap is exactly as described: the builder pins each plugin `.pm` by SHA-256 *because* VEP's header does not name the plugin code — but those checks only run when the builder runs. `is_complete()` accepted any output with 38 plugin fields. The builder now writes the hashes beside the output and the generator rebuilds when they are missing or differ, reusing the pins already in the builder rather than duplicating them.

**P2 — plugin profiles pointed at a file nothing creates (`profiles.py:126`).** Confirmed. The generator writes per-contig files under `output/116/plugins/`; the profile declared a single WGS basename under `output/116/`. It went unnoticed because the runner always passes `--vep` explicitly. Resolution now knows the per-contig template. A multi-contig run has no single file to slice, so it now says exactly that and names both remedies instead of reporting the profile missing.

**P2 — `&`-list order + format together (`compare.py:478`).** Confirmed: `0.10&0.20` vs `0.2&0.1` was recorded as a real value mismatch. Tokens are now paired as a multiset when the position-wise pass fails. The order check still runs first, so a pure reordering stays order-only; exactly one bucket is incremented either way, so the gate's `mismatches + order_only + format_only == strict_unequal` invariant is unchanged. Tests cover the negatives too — differing counts, duplicates that cannot pair one-to-one, and genuinely different values all stay real mismatches.

**P2 — `merged_plugins_base` bypassed the gate guard (`verify_parity_gate.py:65`).** Confirmed, and worth calling out because **this reverses an existing test.** `test_gate_still_accepts_the_plugin_free_base_profile` asserted the gate accepts that profile. But it reads the five-plugin reference, so its report always lists that reference's plugin fields in `fields_only_in_vep`, which the gate rejects at line 413 — gating on it could never pass, it just failed later with "CSQ fields exist only in VEP", reading as a parity defect rather than the wrong profile. The old test encoded a real but subtly wrong premise: attaching no plugins is not the same as comparing against a plugin-free reference.

I have flipped that test rather than worked around it, and said so in the commit. If the intent was that `merged_plugins_base` remain gateable, this is the finding to push back on — say so and I will revert it.

Full suite: 1146 passed, 2 skipped. `ruff check` clean; all shell scripts pass `bash -n`. The one remaining `ruff format` complaint is `performance-tests/vep/scripts/plot_vep_fork_scaling.py`, pre-existing from #44 on master and not in this diff.

@codex pls review
@claude pls review

mwiewior:
Round seven: pushed `43c9126` and `17f2ed6`.

**P2 — the contig never reached resolution (`profiles.py:368`).** Correct, and this was a **regression I introduced last round**, not a pre-existing issue. `resolve()` grew a `chrom` parameter and `main()` never forwarded it, so `chrom` was always `None`, the per-contig lookup was unreachable from the CLI, and every plugin run was rejected — including `--chroms 22`, the exact remedy my own error message recommended. Adding a parameter without wiring it through is squarely on me.

Fixed by passing the sole normalized contig. Two tests cover the wiring specifically, because the defect was that the path could not be *reached*, not that it computed the wrong answer — a test of `resolve()` alone would have passed throughout.

**P2 — references not bound to source data (`generate_vep_plugin_references.sh:47`).** Correct. The sidecar I added recorded only `.pm` hashes, and the reference bytes depend on the plugin data just as much. ClinVar is the clear case: rolling content, fixed filename.

Each of the six sources is now identified by what the server reports (ETag when present, else Last-Modified + Content-Length), recorded beside the `.pm` hashes and revalidated on reuse. The identity helper is shared by the writer and the checker so they cannot drift. Verified against a local server: rewriting ClinVar under the same name changes its recorded identity and triggers a rebuild.

An unreachable source server warns and falls back to validating plugin code alone, rather than forcing a full rebuild of all 22 references over a transient network failure. Say the word if you'd rather that be fatal.

One implementation note: the helper sits beside the scripts rather than under `scripts/lib/`, because `.gitignore`'s packaging-wide `lib/` rule silently excluded it — the first commit attempt added nothing. A shared file that isn't committed is worse than no sharing.

**On the previous round's hang analysis:** your point that the failure needs `$pid` to be the last group member at exactly the poll moment — timing-dependent and easily missed in a handful of runs — is well taken. I reported "does not reproduce", which was true of my harness but is weaker evidence than it sounded. Your reading that `e9e453a`'s re-arm without `break` made it genuinely unbounded is right, and that is now gone.

Full suite: 1148 passed, 2 skipped. `ruff check` clean; all shell scripts pass `bash -n`.

@codex pls review
@claude pls review

mwiewior:
Good catch — real, and pushed as `394c78f`.

You identified the exact gap my verification missed. I had tested the two clean cases (all six sources present, and a server fully down, which fails on the first request before printing anything) and never the flaky middle. Reproduced it against a server holding two of the six:

- **before:** `rc=1`, but **2 `SOURCE` lines already on stdout** — concatenated with the fallback's dump of the old sidecar, giving duplicate conflicting entries for the paths that succeeded
- **after:** `rc=1`, nothing emitted; a complete corpus still yields all six

`all_source_identities()` now buffers and emits only once every source is identified.

Your severity read is right too: the failure direction is a wasted rebuild, never a false "still valid". That is why it survived — it degrades toward caution, so nothing downstream would have complained.

Thanks also for confirming the contig-forwarding fix reaches the right call site with the already-normalized form, and that the `lib/` pattern at `.gitignore:17` is unanchored — I found that one the hard way when the commit added nothing.

Full suite: 1148 passed, 2 skipped. `ruff check` clean; all shell scripts pass `bash -n`.

@codex pls review
@claude pls review

mwiewior:
Round eight: pushed `8510bbb`, `4441c7f`, `12ae753`. Four of the five findings were live; the fifth was already fixed.

**Already fixed — buffer source identities (`generate_vep_plugin_references.sh:38`).** Same issue Claude raised against `17f2ed6`; `394c78f` landed the all-or-nothing buffering before this review ran. Verified against a server holding two of six sources: previously two lines leaked with `rc=1`, now nothing is emitted.

**P2 — reference-free modes blocked (`profiles.py:368`).** Correct. My per-contig rejection was unconditional, so `--skip-annotate` aggregation of stored reports — which opens no reference at all — was refused for want of something it never reads. Now raises only when a reference is required. The documented example gains a contig, and the plugin note says why.

**P2 — runner never validated reference provenance (`run_all_plugin_comparisons.sh:370`).** Correct, and a gap in my own round-six work: I added the sidecar and then never read it from the path that most needs it. Validated before comparing now. All four branches tested: a missing sidecar (the 22 existing references) and an unreachable server both warn and proceed — blocking those would have broken the working pipeline — while a rolling ClinVar change under the same filename is detected and rejected, overridable with `VEP_ALLOW_STALE_REFERENCE=1`.

**P2 — slice reuse ignored source identity (`build_vep_plugin_reference.sh:51`).** Correct, and the sharpest of the four: reusing a stale slice while `plugin_provenance()` recorded identities from the *current* server made the sidecar a false witness — vouching for data the reference never saw. Worse than having no sidecar. Each slice now records the identity it was cut from and is re-sliced when that no longer matches.

**P2 — canonicalisation stopped at the immediate parent (`build_vep_plugin_reference.sh:260`).** Correct. With several missing components under a symlink leaving `$DATA`, the parent is not a directory either, so my fallback returned the lexical path and the containment test accepted it. Now walks up to the longest existing ancestor. Verified: `$DATA/link/new/sub` with `link -> /scratch` is rejected, while nested paths inside `$DATA` still resolve whether or not they exist yet.

One bug of my own, caught by writing the test for the provenance check rather than by review: `local reference="$1" sidecar="$reference.plugins"` cannot cross-reference within a single `local` under `set -u`, which this script runs with. It would have failed on the first call. Split, and I grepped the other functions added in this thread for the same pattern — none.

Full suite: 1149 passed, 2 skipped. `ruff check` clean; all shell scripts pass `bash -n`.

@codex pls review
@claude pls review

mwiewior:
Round nine: pushed `5863da1` and `4ddcd5c`. All three findings were real, and all three were against code added earlier in this thread.

**P2 — reused shards never checked for currency (`:203`).** Correct, and the asymmetry is exactly as described: last round I made the runner validate the *golden reference* against the current sources and never asked the same of our own cache. A chromosome whose shards predate a rolling ClinVar took `CACHE_REUSE`, and the comparison would report stale-cache data as a vepyr mismatch. Shards now record what they were built from; a record that no longer matches forces that plugin to be rebuilt.

**P2 — only the SOURCE half of the sidecar was validated (`:196`).** Correct. `reference_sources_current()` ignored the `PLUGIN` hashes sitting right beside the records it diffed, so a reference built with the wrong `CADD.pm` was accepted whenever the corpus happened to be unchanged — an artifact the generator refuses. Both halves are checked now.

**P2 — slice reuse on an unverifiable identity (`build:62`).** Correct, and the sharpest: my fallback reused the slice whenever HEAD failed, so a source that rolled while connectivity was briefly down would recover in time for `plugin_provenance()` to stamp the *current* identity onto output cut from a stale slice — recreating the false witness the record exists to prevent. No `SOURCE_BASE` at all is a declared mode and still reuses; a configured server that failed to answer now fails with the reason.

Absent records still mean "cannot tell" throughout — the published 97 GiB cache and the 22 existing references predate all of this and must not be invalidated by machinery added after they were built. Verified across all four branches for both shards and references: no record, current, rolled source, unreachable server.

Worth recording two process notes from this round, both mine:

- Moving the shard functions above `cache_complete` (they are called at load time by the source preflight) cut a closing brace. **`bash -n` reported success** — the braces still balanced overall, with `cache_complete` swallowed into `shard_present`. Only the functional test caught it. `bash -n` is a parse check, not a structure check, and I had been leaning on it as though it were both.
- The same edit left a stray `}` behind. I now count `^[a-z_]*() {$` against `^}$` as a cheap structural assertion.

Full suite: 1149 passed, 2 skipped. `ruff check` clean.

On the shape of this: rounds seven through nine have found issues almost entirely in machinery this thread introduced, each smaller than the last. That is convergence, but the provenance system is now a fair amount of new surface for a comparison harness. If you would rather cap it here, `4ddcd5c` is a coherent stopping point — the alternative is that I keep hardening code whose only purpose is to check other code.

@codex pls review
@claude pls review

Issue #7, opened 2026-04-01
Cap tokio blocking thread pool with max_blocking_threads(partitions)


mwiewior:
## Problem

`build_cache(partitions=8)` still consumes ~1480% CPU on a 14-core machine despite all parallelism knobs being threaded through (#73, #74 in datafusion-bio-functions).

## Root Cause

The tokio runtime in `_core.rs` sets `.worker_threads(partitions)` but **not** `.max_blocking_threads()`:

```rust
let rt = tokio::runtime::Builder::new_multi_thread()
    .worker_threads(partitions)
    .enable_all()
    .build()?;
```

Noodles' async BGZF reader calls `tokio::task::spawn_blocking()` for **every** block decompression ([noodles-bgzf/src/async/io/reader/inflate.rs:24](https://github.com/biodatageeks/noodles/blob/9b7b2c5b/noodles-bgzf/src/async/io/reader/inflate.rs#L24)):

```rust
handle: tokio::task::spawn_blocking(move || inflate(buf)),
```

The blocking thread pool is **unbounded by default**. With 8 partition streams each submitting concurrent BGZF decompression tasks, the pool spawns as many OS threads as the OS allows.

## Fix

```rust
let rt = tokio::runtime::Builder::new_multi_thread()
    .worker_threads(partitions)
    .max_blocking_threads(partitions)  // ← ADD THIS
    .enable_all()
    .build()?;
```

This caps total threads to `partitions` async workers + `partitions` blocking workers. The async workers mostly just poll futures (low CPU), while the blocking workers do the actual BGZF decompression (high CPU). Net result: ~`partitions` cores of CPU usage.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #6, opened 2026-03-26
feat: migrate cache building to upstream CacheBuilder

mwiewior:
## Summary
- Replace local `convert.rs` (745 lines) with upstream `CacheBuilder` from biodatageeks/datafusion-bio-functions#67
- Add fjall KV store generation (`build_fjall`, `fjall_zstd_level`, `fjall_dict_size_kb` params)
- Add progress callback support (`on_progress` callable, auto-tqdm bars)
- Remove `memory_limit_gb` parameter (no longer needed)
- Add checked-in Ensembl cache fixture (~24KB, chr22:16M-17M from release 115)
- Add 80 unit + integration tests covering all 6 entities with exact row counts, column schemas, and content validation

## Upstream dependency
- `datafusion-bio-function-vep` pinned to rev `f8e7163` (PR biodatageeks/datafusion-bio-functions#67)

## Test plan
- [x] `pytest tests/test_build_cache.py` — 80 tests, all passing
- [x] Full suite `pytest` — 165+ tests passing
- [x] `cargo clippy` — no warnings
- [x] `ruff check` / `ruff format` — clean

Closes biodatageeks/vepyr#5

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #49, opened 2026-08-24
Multi-part plugin sources + cache-rebuild notebook for bio-functions #217

mwiewior:
### Multi-part plugin source overrides

`build_plugin_cache` now accepts either a path for a single-source manifest or a `{part: path}` mapping for a multi-part manifest:

```python
vepyr.build_plugin_cache(
    "cadd",
    ref,
    source_path={
        "snv": ".../whole_genome_SNVs.tsv.gz",
        "indel": ".../gnomad.genomes.r4.0.indel.tsv.gz",
    },
    ...,
)
```

The mapping must cover every declared part and contain no unknown part. Bare paths remain backward-compatible for single-source manifests and fail clearly for multi-part manifests.

### Real-source cache-build notebook

The notebook builds all five plugin caches chromosome by chromosome and validates the invariants required by the plugin lookup path. Indexed TSV sources are passed as their canonical full BGZF+tabix files so native Rust selection owns chromosome extraction. VCF sources retain the current temporary per-chromosome slicing until the existing VCF provider receives explicit region plumbing.

### Repo hygiene

`23397f0` added about 7 MiB of root-level `.tbi` blobs with no matching source files. Nothing references them — the only mentions of those filenames are the notebook's download lists, which are path-prefixed per plugin and land in `plugin_input/<plugin>/` beside their sources. They are removed here, with a root-anchored `/*.tbi` ignore so it cannot recur. The two legitimate fixtures under `tests/data/` are unaffected.

## Rebased onto master after #48

#48 merged as `6b189e6`, which carried the same `datafusion-bio-function-vep` v0.19.0 pin (`01fa21f`) and the same bio-formats revision (`80f3a6c`). Rebasing resolved every `Cargo.toml`/`Cargo.lock` conflict to that shared final state, so **Cargo no longer appears in this diff at all** — the pins are inherited from master rather than restated here. The branch is now purely its own contribution: the notebook, the multi-part `source_path` API, and the repo-hygiene commit.

Plugin manifests are pinned in the notebook to `biodatageeks/vepyr-plugins@8af8353899ea44b36c80783b781f8bfb5913b242`, the master commit that merged [#7](https://github.com/biodatageeks/vepyr-plugins/pull/7) with [#2](https://github.com/biodatageeks/vepyr-plugins/pull/2) and [#6](https://github.com/biodatageeks/vepyr-plugins/pull/6) already in its ancestry. `build_plugin_cache(version=...)` resolves the ref through a real Git worktree, so only a pushed commit is reachable.

### Local

Re-run after the rebase onto `6b189e6`:

- `cargo fmt --check` clean; `cargo clippy --all-targets -- -D warnings` clean
- `cargo test --no-default-features` — 4 passed
- `pytest` — **1155 passed, 2 skipped** (this branch now carries #48's suite as well as its own)
- one effective bio-functions source (`01fa21f`) and one bio-formats source (`80f3a6c`) in the lockfile

### Native indexed cache build

AlphaMissense chr22 rebuilt at four target partitions from the canonical full 628 MB BGZF + tabix, into an isolated cache root:

- manifest `rows=1,466,988  warm=2,086  cold=1,464,902` — matches the validated shard
- schema identical; `EXCEPT ALL` returns zero rows in **both** directions
- in fact byte-identical to the validated `alphamissense/chr22.parquet` (`cc67264ae74cca466efc3360e12ef8f3`)
- 1.2 s total (0.7 s to read 1.47 M rows), which is only possible with native tabix region selection — a whole-file decompress of the 628 MB source could not complete in that time

### Cache builds

All five release-mode caches were built at four target partitions from the canonical sources; each contains `chr1.parquet` … `chr22.parquet` plus `manifest.json`.

| plugin | shards | rows | bytes |
|---|---:|---:|---:|
| AlphaMissense | 22 | 68,241,195 | 571,181,721 |
| CADD | 22 | 8,370,275,337 | 74,282,931,569 |
| ClinVar | 22 | 4,245,552 | 81,026,132 |
| dbNSFP | 22 | 82,971,900 | 4,030,323,341 |
| SpliceAI | 22 | 3,254,088,469 | 25,478,657,941 |
| **total** | **110** | **11,779,822,453** | **~97.3 GiB** |

All five complete caches were checksum-verified after upload.

### Parity against Ensembl VEP 116

Five-plugin references, chr1–22, engine `01fa21f`:

| | |
|---|---|
| VCF records | 4,096,123 |
| CSQ entries | 69,299,753 |
| strict body MD5 | **22/22 `ok`**, 22/22 exit 0 |
| fields at 100% | **124/124 on every chromosome** |

Zero on all 22, in every category: variants present on only one side, CSQ count / set / order mismatches, value mismatches, format-only mismatches, field-order mismatches, and mismatch-ledger rows.

Headers retain expected run-specific provenance differences; record bodies are byte-identical.

**Scope: autosomes only.** chrX, chrY and chrMT are not covered by these caches or references.


mwiewior:
Rebased onto `master` after #48 merged as `6b189e6`, and force-pushed (`d8f5691` → `8423409`).

**Cargo dropped out of the diff entirely.** #48 carried the identical v0.19.0 pin (`01fa21f`) and bio-formats revision (`80f3a6c`), so every `Cargo.toml`/`Cargo.lock` conflict resolved to that shared final state and the files are now unchanged relative to master. The net diff is 6 files — the notebook, the multi-part `source_path` API in `src/lib.rs` / `__init__.py` / `_core.pyi`, its tests, and the `.gitignore` guard.

The eight commits replay cleanly with none empty. One cosmetic artifact of the rebase: `e6cb951` still reads "pin the engine build that fixes the CADD tier-join overcount" but now contains only the orphan `.tbi` blobs, since its Cargo change was subsumed by master. `8423409` removes those blobs again, so the net tree has none — and this repo squash-merges, so neither the message nor the blobs reach `master`.

Gate re-run after the rebase: `cargo fmt --check` clean, `clippy --all-targets -D warnings` clean, 4 Rust tests, **1155 passed / 2 skipped** (this branch now carries #48's suite too). One effective bio-functions source and one bio-formats source in the lockfile.

@codex pls review
@claude pls review

mwiewior:
Both findings fixed in `2999907`.

**P1 — `validate()` materialized whole shards (`notebook:424`).** Correct, and worse than a bound being exceeded: it could never have run. `build_genome()` calls `validate()` for every chromosome, and **CADD chr1 is 699,768,898 rows** — `to_pylist()` on that plus a set of one tuple per row is far beyond the 64 GiB host, so the notebook's claimed 3–4 GB bound and "verified exhaustively" were not reachable together at genome scale.

Now streams fixed-size batches. Ordering runs as Arrow compute per batch, carrying the last row across the boundary. Duplicate detection uses the sorted contract the same function asserts — rows sharing a `(tier, start)` are contiguous, so a duplicate can only occur inside one run, and carrying the trailing run across a batch boundary is enough.

Measured against the real caches, peak RSS rather than an estimate:

| shard | rows | time | peak RSS |
|---|---:|---:|---:|
| clinvar chr22 | 96,223 | <1s | 58 MB |
| spliceai chr22 | 52,715,028 | 5.7s | 187 MB |
| cadd chr22 | 119,170,699 | 10.6s | 190 MB |
| **cadd chr1** | **699,768,898** | **62s** | **162 MB** |

All 699,768,898 CADD chr1 keys verified exhaustively at 162 MB.

**P2 — the plugin guide contradicted the new API (`docs/plugins.md:78-81`).** Correct: it still said `build_plugin_cache()` takes a single `source_path` and rejects manifests with more than one `[[source]]`, so a reader would conclude this PR's own CADD workflow was impossible. Rewritten with the mapping shape and CADD as the worked example.

While writing it I nearly asserted a second rejection direction — a mapping against a single-source manifest — from a reading of the code. Only four directions had tests, and that was not one of them. Checked it instead: it is rejected, with `source_path was given as a dict, but a [[source]] in plugin "demo" declares no `part` to key it by`. The doc now states that reason, and a test covers it.

Full suite: 1156 passed, 2 skipped. `ruff check` and `ruff format` clean.

Separately, the full chr1–22 five-plugin parity run is in flight on this branch with the extension rebuilt from it. chr22 is already byte-identical to the release-116 reference (`8c6ae1e17452776316ab3ee48def0ebf`), zero mismatches, 124/124 fields.

@codex pls review
@claude pls review

mwiewior:
All three fixed in `5db47b6`. One of them was a regression in my own previous fix.

**Cross-tier duplicate blind spot (@claude).** You were right, and I verified the premise before fixing rather than after: **1,117,083 cold rows of CADD chr22 share a `start` with a warm row**. So the same `start` really does occur in both blocks, my `(tier, start)` segmentation never compared those pairs, and the pre-streaming version did not have that hole. I traded memory for correctness, which was not the trade on offer.

Now covers all three cases — warm×warm against a key set small enough to hold, cold×cold against contiguous runs in the start-sorted cold block, and warm×cold only where a cold start also occurs in warm, which is a small filtered subset. Verified on synthetic shards:

| case | expected | result |
|---|---|---|
| clean shard, same start in both tiers, different alleles | accept | accepted |
| duplicate straddling the tier boundary | detect | **detected** |
| duplicate inside the cold block | detect | detected |
| duplicate inside the warm block | detect | detected |

Cost, stated plainly: CADD chr1 goes 62s → 366s (976 MB peak). The 62s version was wrong.

To your question of whether tiering is deterministic per full key — I did not assume either way. The check no longer depends on it.

**Header-only slices (@codex).** Correct, and the masking was exactly as described. `sh()` used `shell=True` under `/bin/sh`, so `tabix … | bgzip -c > out` reported bgzip's status; `-h` then made the file non-empty and `validate()` accepted a zero-row shard, because ordering and uniqueness hold vacuously over no rows. `sh()` now runs under bash with `pipefail` — `/bin/sh` is not enough, since it is dash on many Linux distros and dash has no `pipefail` — and a slice must contain at least one data record. Verified: `false | true` now fails, legitimate pipelines still pass, and a header-only bgzip file is rejected where the size check passed it.

**Resume space check (@codex).** Correct. Measured, fresh vs resume: **222.0 GB → 222.0 GB** (unchanged) and **222.0 GB → 133.2 GB**. The documented host has ~145 GB free once CADD has landed, so the resume path now fits where it previously aborted before doing any work.

Full suite: 1156 passed, 2 skipped. `ruff check`/`ruff format` clean.

Parity meanwhile: 6/22 chromosomes complete on this branch with the extension rebuilt from it, every one byte-identical to the pre-merge run, zero mismatches, 124/124 fields.

@codex pls review
@claude pls review

mwiewior:
Fixed in `9eaa3dc` — and correct: this was a hole in the previous commit's own fix. The record check ran only on the creation path, so a header-only slice left by a failed run was returned by `if out.exists()` before the check was ever reached. Reuse now clears the same bar as creation, and a slice is written under a partial name and published only once it has records, so a failure cannot leave an invalid file for the next run to find. Verified: a stale header-only slice is detected, discarded and re-cut.

**Closing further notebook review here**, deliberately, and it is worth saying why rather than just going quiet.

`notebooks/build_plugin_caches.ipynb` is an operational driver, not shipped library code. The cache build itself is owned by bio-functions `v0.19.0`, pinned from `master`; the notebook fetches sources, cuts temporary slices for the two plugins whose VCF provider still lacks region plumbing, calls `build_plugin_cache()`, and asserts over the result. This PR's actual product change is `src/lib.rs` (+91) — the multi-part `source_path` mapping — plus its 17 tests, and both reviewers have read that clean since the first pass.

The notebook's `validate()` checks (a) the engine's own `ORDER BY tier, start` storage contract, and (b) `assume_unique`. Both are already covered, more convincingly, by the full chr1–22 parity gate: if either were violated the output would diverge from Ensembl VEP. The one thing parity genuinely cannot reach is cache rows no test VCF touches — HG002 hits ~4M positions of 11.8B cached rows — which is real but narrow. Against that narrow value, the check has now needed correcting twice in two rounds (materialisation, then a cross-tier blind spot). Continuing to harden it is a poor trade, so we are stopping.

Findings on `src/lib.rs`, the Python API, its tests, or `docs/plugins.md` remain welcome and will be acted on.

Parity status on this branch, extension rebuilt from it: **7 of 22 chromosomes complete, every one byte-identical to the pre-merge run** — zero variants on one side only, zero CSQ count/set/order mismatches, zero value/format/field-order mismatches, empty ledgers, 124/124 fields. The remaining chromosomes are running; the full result will be posted when it finishes.

PR #64, opened 2026-09-03
chore: qualify the release-116 cache-build determinism fixes

mwiewior:
Pins the three upstream fixes for the cache-build non-determinism reported in [sitekwb/vepyr-porting-tests#387](https://github.com/sitekwb/vepyr-porting-tests/issues/387), and updates the schema docs and tests that the column change invalidates.

All upstream work is **merged**; this PR carries the merged master SHAs.

## The defects

Building the same Parquet cache twice, from the same raw cache with the same command, did not give the same bytes. Two independent causes:

- **`translation_core` was reordered** — its export query was the only one in the cache builder without an outer `ORDER BY`. `execute_stream` wraps a multi-partition plan in `CoalescePartitionsExec`, an *arrival-ordered* merge, so the shard was N internally-sorted runs concatenated in whatever order the threads finished. Measured at `partitions=8`: chr21 88% of positions moved, chr22 100% — a pure permutation (same multiset, equal once sorted).
- **`exon` picked a different source copy** — the dedup ordered only by `stable_id`, which is *identical* in the two cross-region copies of an exon whose transcript straddles a 1 Mb cache-region boundary. A complete tie, so the winner was whichever row reached the window operator first. Not merely non-deterministic: transcript and translation both break this tie explicitly, citing VEP's `merge_features` "lowest region wins". Exon was never given that rule, making it a latent correctness bug, not just a reproducibility one.

While tracing which columns the exon race could touch, it turned out `raw_object_json`/`object_hash` are read by nobody — they appear in none of the `*_CONTEXT_COLUMNS` projections and are never read by name. The exon shard was additionally leaking `_rn`, the dedup helper.

## Pins

| rev | contents |
|---|---|
| bio-formats `12016a9` | v1.12.0 + biodatageeks/datafusion-bio-formats#246 (exon dedup tie-break) |
| bio-functions `86f1c0a` | v0.19.2 + #224 (translation_core `ORDER BY` + CI step) + #225 (drop unread columns) + #226 (bio-formats pin bump) |

Both crates carry the same bio-formats rev, which is required and not incidental: `bio-function-vep` pins its own, and two revs are two distinct sources to cargo, so a mismatch compiles the formats crates twice and `CacheSourceType` fails to cross into `CacheBuilder` (E0308). `Cargo.lock` resolves to `12016a9` x4 and `86f1c0a` x1.

## Qualification

Rebuilt all three release-116 caches (ensembl, merged, refseq). Only **5 of 7 entities** needed it — `variation` (29.2 GB) and `translation_sift` (4.4 GB) use different write paths, which cut the rebuild from ~5.5 h to ~2 h. That scoping was verified, not assumed: `variation/chr22` rebuilt under the new code is **byte-identical** to the pre-change shard (388,756,098 bytes, `cmp` clean), and `translation_sift` came back byte-identical on 40/40 sampled shards.

**Cache size**, per cache (ensembl shown; refseq/merged comparable):

| entity | before | after | |
|---|---:|---:|---:|
| transcript | 659.6 M | 276.6 M | -58.1% |
| exon | 218.2 M | 128.4 M | -41.2% |
| regulatory | 74.1 M | 7.6 M | -89.7% |
| motif | 92.8 M | 28.5 M | -69.3% |
| translation_core | 272.5 M | 219.1 M | -19.6% |
| translation_sift | 4359.4 M | 4359.4 M | byte-identical |
| variation | 29171.3 M | 29171.3 M | untouched |
| **total** | **34.85 G** | **34.19 G** | **-660 MB** |

`translation_core` shrinking 19.6% was not predicted — that is the `ORDER BY` improving dictionary encoding. The determinism fix pays for itself in size.

**Determinism**: an independent rebuild of exon chr21+chr22 at `partitions=8` is **byte-identical** to the cache shard. Pre-fix, chr22 varied in roughly 1 of 8 builds (measured: 8 builds produced 2 distinct shards). Re-confirmed after re-pinning to the merged SHAs.

**Annotation parity** — HG002 WGS re-annotated against all three rebuilt caches, md5 vs the Ensembl VEP 116 references:

| profile | contigs | strict | canonical | records |
|---|---|---|---|---:|
| ensembl | 24 | PASS 24/24 | PASS 24/24 | 4,317,549 |
| refseq | 24 | PASS 24/24 | PASS 24/24 | 4,317,549 |
| merged | 22 | PASS 22/22 | PASS 22/22 | 4,096,123 |

**140/140 body digests match, zero failures.** `HEADER` shows `DIFF` on every row by design — VEP stamps run-specific provenance (`##VEP=` timestamp), which the tool hashes separately.

Two limits stated plainly rather than glossed:
- **merged is 22 contigs** because its VEP reference has no chrX/chrY records, not a partial run.
- **`merged_plugins` was not exercised** — no plugin cache is present locally for ClinVar/SpliceAI/CADD/AlphaMissense/dbNSFP — so it is excluded rather than counted as passing.

## Squash-merge verification

All three upstream PRs were squash-merged, so the revs the caches were qualified against are **not** ancestors of master and ancestry proves nothing. Content compared instead:

```
bio-formats   export_query.rs   c992f69 vs 12016a9  ->  identical
bio-functions cache/build.rs    93ba78e vs 86f1c0a  ->  identical
```

So the qualification above carries over unchanged, with no repeat of the 2-hour rebuild.

## Docs and tests

Four entity tables in `docs/caches.md` lose `raw_object_json`/`object_hash` (exon also `_rn`); column counts corrected (transcript 59 to 57, exon 16 to 13, regulatory 12 to 10, motif 17 to 15), each verified against its actual row count.

Three `*_column_count` tests failed on the new schema — real signal, asserting 78/35/31 against the actual 76/32/29. Updated, plus a new `*_omits_unpersisted_columns` test per entity that names the removed columns rather than trusting a bare count to catch a regression.

`uv run pytest`: **1174 passed, 2 skipped**.

## Merge path (complete)

1. bio-formats #246 — merged `12016a9`
2. bio-functions #224 — merged `ca3ef85`
3. bio-functions #225 — merged `213f6ff` (rebased onto #224; both add tests at the same anchor)
4. bio-functions #226 — merged `86f1c0a` (bio-formats pin bump)
5. **this PR** — re-pinned to the merged SHAs
6. `integration/387-cache-build-fixes` — deleted

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1150,8 +1150,8 @@ dependencies = [
 
 [[package]]
 name = "datafusion-bio-format-bed"
-version = "1.11.0"
-source = "git+https://github.com/biodatageeks/datafusion-bio-formats.git?rev=80f3a6c043899f8ded44eaab204b42735f805272#80f3a6c043899f8ded44eaab204b42735f805272"
+version = "1.12.0"
+source = "git+https://github.com/biodatageeks/datafusion-bio-formats.git?rev=c992f69e0b94201c4d9056b7450f2b2ec2922d5e#c992f69e0b94201c4d9056b7450f2b2ec2922d5e"
 dependencies = [
  "async-compression",
  "async-stream",
@@ -1173,8 +1173,8 @@ dependencies = [
 
 [[package]]
 name = "datafusion-bio-format-core"
-version = "1.11.0"
-source = "git+https://github.com/biodatageeks/datafusion-bio-formats.git?rev=80f3a6c043899f8ded44eaab204b42735f805272#80f3a6c043899f8ded44eaab204b42735f805272"
+version = "1.12.0"
+source = "git+https://github.com/biodatageeks/datafusion-bio-formats.git?rev=c992f69e0b94201c4d9056b7450f2b2ec2922d5e#c992f69e0b94201c4d9056b7450f2b2ec2922d5e"
 dependencies = [
  "async-compression",
  "bytes",
@@ -1196,8 +1196,8 @@ dependencies = [
 
 [[package]]
 name = "datafusion-bio-format-ensembl-cache"
-version = "1.11.0"
-source = "git+https://github.com/biodatageeks/datafusion-bio-formats.git?rev=80f3a6c043899f8ded44eaab204b42735f805272#80f3a6c043899f8ded44eaab204b42735f805272"
+version = "1.12.0"
+source = "git+https://github.com/biodatageeks/datafusion-bio-formats.git?rev=c992f69e0b94201c4d9056b7450f2b2ec2922d5e#c992f69e0b94201c4d9056b7450f2b2ec2922d5e"
 dependencies = [
  "async-trait",
  "datafusion",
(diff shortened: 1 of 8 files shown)


mwiewior:
@claude review

mwiewior:
@codex review

mwiewior:
@codex review

mwiewior:
@codex review

mwiewior:
@claude review

PR #73, opened 2026-09-04
feat: record plugin cache source revisions

### LINKED ISSUE #70: Plugin cache manifest.json never records the vepyr-plugins ref it was built from (cache_source_version is always null)
## Summary

Every plugin cache's `plugin/<name>/manifest.json` carries a `cache_source_version` field, but the builder always writes it as `null`. A built plugin cache therefore does not record which `vepyr-plugins` git ref (tag / commit) its source manifest came from, even though `build_plugin_cache()` is given that ref explicitly as `version`.

## Where it happens

- `datafusion-bio-function-vep/src/plugin_cache/cache_manifest.rs` (rev `fcadc1b`, currently pinned in `Cargo.toml`): `CacheManifest::from_source()` hard-codes `cache_source_version: None`, and nothing on `PluginCacheBuilder` (`with_chrom_filter`, `with_overwrite`, `with_source_verification`) can set it afterwards.
- `src/lib.rs` `build_plugin_cache` (pyo3): takes `manifest_path` but not the resolved ref, so it cannot pass anything down even if the builder accepted it.
- `src/vepyr/__init__.py` `_resolve_plugin_manifest()`: checks out `version` via `git worktree add --detach`, so the exact commit is known at this point and then discarded.

Only the manifest *file name* survives into the cache (`source_manifest: "alphamissense.source.toml"`).

## Why it matters

- The `sources` block added in #9 / 9e40f40 now records `url` / `md5` / `verified_md5` per input file, so the **data** provenance is traceable, but the **manifest** provenance (which `ingest_sql`, `value_columns`, `csq_rank`, `field_order`, `allele_match` the cache was built with) is not. Two caches built from the same data at different `vepyr-plugins` tags can emit different CSQ fields and are indistinguishable from their `manifest.json`.
- Prebuilt caches published on Hugging Face cannot state which manifest release they correspond to.
- `annotate(expected_cache_version=...)` exists for the Ensembl cache; there is no equivalent assertion possible for plugin caches.
- Incremental builds (`chroms=[...]`) upsert into an existing `manifest.json`; without a recorded ref there is no way to detect that shards for different chromosomes were built from different manifest revisions.

## Proposal

1. `bio-functions`: add `PluginCacheBuilder::with_source_version(impl Into<String>)` that populates `cache_source_version` (keep `None` when not supplied so existing callers are unaffected). On an incremental upsert, if the seeded manifest already has a different non-null `cache_source_version`, fail (or warn) rather than silently mixing revisions.
2. `vepyr` `src/lib.rs`: accept an optional `source_version: Option<&str>` on `build_plugin_cache` and forward it.
3. `vepyr` Python: in `_resolve_plugin_manifest()` also yield the resolved commit (`git -C <worktree> rev-parse HEAD`) and pass `"<version>@<sha>"` (or the tag and sha as two fields) to the engine. Recording the sha as well as the user-supplied ref matters because a branch name like `master` is not stable.
4. Docs: mention the field in the plugin-cache layout section of `docs/plugins.md`.

## Optional follow-up

Once the field is populated, `annotate()` could grow an `expected_plugin_version={"cadd": "v0.2.0", ...}` assertion mirroring `expected_cache_version`.

https://claude.ai/code/session_016sUVsmcgEihuHznA89owyo

mwiewior:
Closes #70.

Records each requested plugin ref together with its resolved immutable commit SHA, forwards that provenance through the PyO3 boundary, and refuses mixed-revision incremental cache updates in the upstream builder. Full overwrite rebuilds may advance revisions.

Depends on biodatageeks/datafusion-bio-functions#231.

Validation:
- 1182 pytest tests passed, 2 skipped
- 18 plugin-cache tests passed
- cargo clippy --no-default-features --all-targets -- -D warnings
- cargo fmt --check

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1246,7 +1246,7 @@ dependencies = [
 [[package]]
 name = "datafusion-bio-function-vep"
 version = "0.17.2"
-source = "git+https://github.com/biodatageeks/datafusion-bio-functions.git?rev=58b5650750db246270dd8bd564505fc080eb8796#58b5650750db246270dd8bd564505fc080eb8796"
+source = "git+https://github.com/biodatageeks/datafusion-bio-functions.git?rev=e8132dba5bbb321e8f14f235ad8405f24c01d1e3#e8132dba5bbb321e8f14f235ad8405f24c01d1e3"
 dependencies = [
  "ahash",
  "async-trait",
@@ -1264,7 +1264,9 @@ dependencies = [
  "indicatif",
  "libc",
  "log",
+ "md-5",
  "nohash-hasher",
+ "noodles-bgzf 0.49.0",
  "noodles-core 0.16.0",
  "noodles-core 0.20.0",
  "noodles-csi",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1246,7 +1246,7 @@ dependencies = [
 [[package]]
 name = "datafusion-bio-function-vep"
 version = "0.17.2"
-source = "git+https://github.com/biodatageeks/datafusion-bio-functions.git?rev=e8132dba5bbb321e8f14f235ad8405f24c01d1e3#e8132dba5bbb321e8f14f235ad8405f24c01d1e3"
+source = "git+https://github.com/biodatageeks/datafusion-bio-functions.git?rev=dc191d078e2c89a91686527440013954217926e0#dc191d078e2c89a91686527440013954217926e0"
 dependencies = [
  "ahash",
  "async-trait",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1246,7 +1246,7 @@ dependencies = [
 [[package]]
 name = "datafusion-bio-function-vep"
 version = "0.17.2"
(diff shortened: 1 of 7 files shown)


mwiewior:
@codex review

mwiewior:
@claude review

mwiewior:
@claude review

mwiewior:
@codex review

mwiewior:
@claude review

mwiewior:
@codex review

mwiewior:
@claude review

mwiewior:
@codex review\n\nThe head now pins the reviewed engine fix at 2789482.

mwiewior:
@codex review

Remote-only plugin branches now resolve through origin/<version> to an immutable commit before worktree creation, with a fresh-clone regression test.

mwiewior:
@claude review

Addressed the remote-branch finding at e75ae75, including a fresh-clone regression test; the engine pin will follow the latest compatibility fix.

mwiewior:
@codex review

Please review current head d12fcf1, which only adds the final reviewed engine compatibility pin after the remote-branch fix.

mwiewior:
@codex review

PR #75, opened 2026-09-04
feat: add plugin-only core field annotation

### LINKED ISSUE #72: Plugin-only annotation: emit VEP's core @DEFAULT_OUTPUT_COLS plus plugin fields (fields="core")
## Summary

Allow vepyr to be used purely as a plugin annotator: emit only VEP's core output columns plus the selected plugins' fields, instead of the full 74-field `CSQ` block that every run produces today.

The core set is Ensembl's `@DEFAULT_OUTPUT_COLS` in `Bio::EnsEMBL::VEP::Constants` (release/116, [Constants.pm#L268-L282](https://github.com/Ensembl/ensembl-vep/blob/release/116/modules/Bio/EnsEMBL/VEP/Constants.pm#L268-L282)):

```
Uploaded_variation  Location  Allele  Gene  Feature  Feature_type  Consequence
cDNA_position  CDS_position  Protein_position  Amino_acids  Codons  Existing_variation
```

This is what `vep` emits with no output flags at all. Note that in VCF output VEP drops `Uploaded_variation` and `Location` from `CSQ` (they are `CHROM`/`POS`/`ID`), so the VCF-side core is the remaining eleven.

## Motivation

- A common workflow already has a full VEP or vepyr annotation and only wants to **add** CADD / SpliceAI / AlphaMissense / dbNSFP / ClinVar scores. Today that means re-emitting (and, for the DataFrame path, re-parsing) the whole 74-field block just to reach the plugin values appended at the end.
- The core columns are exactly the ones plugin matching needs: per-variant plugins key on `Allele`, per-transcript plugins on `Feature` plus `Protein_position` / `Amino_acids` (AlphaMissense, dbNSFP) or `SYMBOL` (SpliceAI, which would need adding to the set for that plugin). Everything else in the 74 fields is dead weight for this use case.
- The base field count currently varies with cache type and `everything` (79 / 85 / 86, see #57), which makes positional access to plugin fields fragile. A fixed, minimal core makes the layout predictable.
- It is the natural counterpart of VEP's `--fields` option, which vepyr does not expose.

## Proposal

1. Add `fields` to `annotate()`, mirroring VEP's `--fields`: either the preset string `"core"` (= `@DEFAULT_OUTPUT_COLS`) or an explicit ordered list of field names. Default stays the current full set, so existing output is unchanged.
2. Plugin-only annotation is then `annotate(..., fields="core", plugin_cache_root=..., plugins=[...])`: the `CSQ` header `Format:` lists the core columns followed by the plugin blocks (in the order given to `plugins`, per #71).
3. Apply the selection in **both** the header pass and the per-transcript body pass in the engine, the same way plugin discovery has to agree between the two (see the note in `_plugin_subset_root`). A field in `fields` that a plugin's match template needs (e.g. `SYMBOL` for SpliceAI) should be computed even if not emitted, or the call should fail with a clear message.
4. Do not compute what is not emitted where it is expensive: HGVS is already opt-in, but this mode should also let the engine skip optional lookups such as `Existing_variation` when it is not in `fields` (ties in with #18, deferring output materialisation).
5. On the DataFrame path, the selected fields and the plugin fields should surface as named columns rather than positions inside `CSQ` (this is the fix #57 asks for; a small fixed field set makes it tractable).

## Acceptance

- `annotate(vcf, cache_dir, fields="core", plugin_cache_root=root, plugins=["cadd"], output_vcf=out)` writes a `CSQ` `Format:` of exactly `Allele|Gene|Feature|Feature_type|Consequence|cDNA_position|CDS_position|Protein_position|Amino_acids|Codons|Existing_variation|CADD_PHRED|CADD_RAW`.
- Byte parity with `vep --fields "<same list>" --plugin CADD,...` on the golden 116 chr22 inputs for each of the five catalogued plugins.
- Runtime and memory for a plugin-only run are measurably below a default run on the same input (record numbers in the PR).

## Related

- #57 plugin fields unusable from the DataFrame path
- #71 plugin block order from `annotate(plugins=[...])`
- #18 defer output materialisation until after pick

https://claude.ai/code/session_016sUVsmcgEihuHznA89owyo

mwiewior:
Closes #72.

Depends on biodatageeks/datafusion-bio-functions#233 and #74.

Adds `fields="core"` and ordered explicit field selections to both VCF and DataFrame annotation. The VCF header/body share the selected layout, plugin fields remain a trailing ordered block, and DataFrames expose only the selected named annotation columns plus named plugin list columns. The engine keeps match-template dependencies available while skipping unselected expensive work and writes the core layout directly.

Validation:
- 1,186 pytest tests passed (2 skipped)
- full engine suite: 1,040 unit tests passed (1 ignored), plus 3 plugin and 4 VCF integration tests
- Rust clippy/fmt and Python Ruff pass
- actual golden VCF/plugin integration verifies the exact header/body layout and named DataFrame plugin output

Release benchmark (Apple arm64, 10,000 repeated golden variants, one demo plugin, 8 alternating runs after warm-up): default median 0.9683 s vs core 0.7464 s (22.9% faster); plain VCF 175,345,115 vs 83,997,833 bytes (52.1% smaller); retained DataFrame size 288,396,304 vs 67,240,550 bytes (76.7% smaller).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1246,7 +1246,7 @@ dependencies = [
 [[package]]
 name = "datafusion-bio-function-vep"
 version = "0.17.2"
-source = "git+https://github.com/biodatageeks/datafusion-bio-functions.git?rev=90ed9b0371b7bdac02b8a1c46bb82d593b1424c4#90ed9b0371b7bdac02b8a1c46bb82d593b1424c4"
+source = "git+https://github.com/biodatageeks/datafusion-bio-functions.git?rev=afff39e0aaae293d2a1b6c6005e279e6e54bcca6#afff39e0aaae293d2a1b6c6005e279e6e54bcca6"
 dependencies = [
  "ahash",
  "async-trait",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1246,7 +1246,7 @@ dependencies = [
 [[package]]
 name = "datafusion-bio-function-vep"
 version = "0.17.2"
-source = "git+https://github.com/biodatageeks/datafusion-bio-functions.git?rev=9157c950ce8191dacbb6982679c405cc2aa5bc8c#9157c950ce8191dacbb6982679c405cc2aa5bc8c"
+source = "git+https://github.com/biodatageeks/datafusion-bio-functions.git?rev=a0298dd122cb98310b2da3294fba700e8238ed91#a0298dd122cb98310b2da3294fba700e8238ed91"
 dependencies = [
  "ahash",
  "async-trait",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1246,7 +1246,7 @@ dependencies = [
 [[package]]
 name = "datafusion-bio-function-vep"
 version = "0.17.2"
-source = "git+https://github.com/biodatageeks/datafusion-bio-functions.git?rev=a0298dd122cb98310b2da3294fba700e8238ed91#a0298dd122cb98310b2da3294fba700e8238ed91"
+source = "git+https://github.com/biodatageeks/datafusion-bio-functions.git?rev=90ed9b0371b7bdac02b8a1c46bb82d593b1424c4#90ed9b0371b7bdac02b8a1c46bb82d593b1424c4"
 dependencies = [
  "ahash",
  "async-trait",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -64,6 +64,6 @@ serde_json = "1"
 # two different revs are two different sources to cargo, so a mismatch compiles
 # the formats crates twice and `CacheSourceType` / the VCF writer types then fail
 # to cross into `CacheBuilder` (E0308).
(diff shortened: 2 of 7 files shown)


mwiewior:
@codex review

mwiewior:
@claude review

mwiewior:
@codex review

Current head 26f796e pins the engine head containing the direct core-writer parity regression.

mwiewior:
@claude review

Please re-review current head 26f796e after the engine parity-test pin update.

mwiewior:
@codex review

Current head a2913cc restricts selected fields to the annotation portion of the schema and adds direct boundary/missing-plugin-directory regressions.

mwiewior:
@claude review

Addressed the actionable review points in a2913cc: fixed-prefix names cannot become duplicate selections, a missing plugin directory gets the established clear error, and the redundant CSQ drop is removed. Both cases have regression tests.

mwiewior:
@codex review

PR #84, opened 2026-09-06
chore: declare the notebook dev dependencies

mwiewior:
## Why

Installing a Jupyter stack into `.venv` with `uv pip install` doesn't survive. `uv sync` prunes anything missing from `uv.lock`, so the next sync — PyCharm's uv integration runs one on project load — deletes `ipykernel`, `jupyterlab`, `ipywidgets` and `jupyter_server`.

Worse, a sync **while a server is running** strands it. The server keeps answering, but every kernel start dies on import, so `POST /api/kernels` returns:

```
{"message": "Kernel does not exist: 590b6a50-1e4c-4651-a526-d02d6f6bc72a"}
```

for an id it just minted. IntelliJ/PyCharm surfaces that as `JupyterRestClientErrorResponseException … code 500, message: HTTP 404: Not Found (Kernel does not exist: …)`, which looks like a stale-session bug and isn't one.

## What

Declare `ipykernel`, `jupyterlab` and `ipywidgets` in the `dev` group, which a bare `uv sync` installs. `ipywidgets` is already user-facing through the `notebook` extra, but extras aren't installed by a bare sync, so the dev group needs its own entry. The `notebook` package itself is deliberately not declared — JupyterLab is what the IDE launches, and it brings `jupyter_server`.

## Verification

Before: a plain `uv sync` left `GONE ipykernel / jupyterlab / notebook / jupyter_server / ipywidgets`.

After, the same `uv sync`:

```
survives sync: ipykernel
survives sync: jupyterlab
survives sync: ipywidgets
survives sync: jupyter_server
survives sync: vepyr
```

And a fresh server from the synced venv starts a kernel that stays up, with a real `ipykernel_launcher` process — the exact operation that was failing.

https://claude.ai/code/session_01CaYGAYHGV5JAkcAaGgo71j

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -36,6 +36,13 @@ dev = [
     # figures for the scaling benchmarks
     # (performance-tests/vepyr/scripts/plot_vepyr_worker_scaling.py).
     "matplotlib>=3.8",
+    # notebooks/annotate.ipynb, and the Jupyter server an IDE starts for it.
+    # These must be declared, not `uv pip install`ed: `uv sync` prunes anything
+    # missing from uv.lock, which strands a running server with no kernels.
+    # ipywidgets is also user-facing via the `notebook` extra (tqdm.auto bars).
+    "ipykernel>=7.3.0",
+    "jupyterlab>=4.6.3",
+    "ipywidgets>=8.1.8",
 ]
 
 [tool.maturin]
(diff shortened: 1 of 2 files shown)


mwiewior:
@codex review

mwiewior:
@claude review

PR #100, opened 2026-09-09
docs(skills): add the vepyr-fix workflow skill

mwiewior:
## What this adds

`.claude/skills/vepyr-fix/SKILL.md` — a repo-level skill that encodes the routine
for landing a vepyr fix across `datafusion-bio-formats`,
`datafusion-bio-functions` and `vepyr`.

It exists because the two things that decide whether a fix landed well are
measured on the developer's own machine rather than argued about: byte-level md5
parity against Ensembl VEP, and timings compared against the same host's earlier
numbers. Both have to be captured **before** any edit for the end-state
comparison to mean anything, and that ordering is easy to get wrong once.

## Shape

Eight steps, each ending in a stated gate: baseline, failing test, implement in
dependency order, one draft PR per repo with the pin cascade, request both
reviewers, iterate to green under a monitor, re-verify, hand off.

**The skill never merges.** Verification runs on the stacked branches and has to
pass before any PR leaves draft, so a human is only ever asked to review work
whose gates already passed. Merging and the post-merge re-pin are left to them.

## Gates

| Gate | Criterion |
|---|---|
| Quality | every md5 body digest matches in `strict` mode, and `verify_parity_gate.py` exits 0 |
| Performance | no `[VEP_PIPELINE_TRACE]` phase regresses >5%, total wall within 10% at 1 and 8 workers |

## Verification of the content itself

Every command is checked against the harness it drives, at the pins master uses
(engine `v0.20.1` = `8b1abd0`, formats `v1.12.1` = `419be98`). Three things the
drafting caught that are worth knowing independently of the skill:

- Measured runs need `RUSTFLAGS="-C target-cpu=native" uv sync --reinstall-package vepyr`,
  per `docs/performance.md`. `maturin develop` is the fast-iteration build, and
  mixing the two across the bookends compares different binaries rather than
  different code.
- `run_comparison.py` takes contig names or `all`; only `verify_parity_gate.py`
  expands `1-22`. Passing a range to the former treats it as one contig name and
  compares nothing.
- `--minimum-free-gib` is on the Python runner, not the shell wrapper, so raising
  the floor from 40 to the ~65 GiB that plain output above one worker actually
  needs means calling `run_vepyr_worker_scaling.py` directly.

It also carries the measurement traps that have each cost a cycle before: the
shared host, the first-run outlier at 8 workers, RSS attribution by ablation
rather than phase boundary, and preferring phase durations to wall clock.

## Notes for review

- Docs-only, so CI has nothing to build here and the pre-commit hooks all skipped.
- Written outside a GSD workflow at the maintainer's explicit request. `.planning/`
  in this repo carries no `ROADMAP.md`, so `/gsd:quick` cannot run.
- Codex has no committed review workflow in any of the three repos — it answers as
  a GitHub App. The skill says so, so a silent codex is diagnosed as an app
  problem rather than a missing workflow file.


mwiewior:
@codex review

mwiewior:
@claude review

mwiewior:
Thanks — all four confirmed against the sources and fixed in fb5ec34..15ed212.

1. **Absolute `RUN`** — correct, the later `cd` broke the redirections. Now resolved from `git rev-parse --show-toplevel`, and every subsequent `cd` is absolute.
2. **`--force` on the measured sweep** — correct, `run_vepyr_worker_scaling.py:289-293` refuses any pre-existing artifact for that worker count, so the measured run died after a successful warm-up. Added `--force`, and the warm-up now writes to its own archive directory so it cannot reach the measured summary.
3. **Trace source** — correct, and this was the important one. Each child's stderr goes to its own artifact and is archived (`:330-333`, `:346`), so the redirect captured only progress chatter and the gate diffed two empty files, reporting no phase regression having measured nothing. Now read from `archive/merged_workers<N>.stderr.txt`, with an explicit non-empty assertion so an empty trace can never read as green.
4. **Report ordering** — correct, md5 mode leaves `comparison` null (`cli.py:418-419`) at the same report path as field mode (`:466`), and the gate rejects that (`verify_parity_gate.py:398-400`). Now field mode, then the gate, then the digests, with the reports copied aside because the md5 pass overwrites them. There is no `--report-dir` to separate them (`cli.py:588`).

The perf step also moves from the shell wrappers to the Python runner, the only entry point that can express separate archive directories and the 65 GiB floor. Every bash block in the file now parses under `bash -n`.

@codex review
@claude review

mwiewior:
Round two — four real, one stale. Fixed in c568707.

**1. `tee` masking failures** — correct, and the most dangerous of the five. `run_comparison.py` returns 1 on any mismatch (`comparison/cli.py:695`) and the gate exits nonzero, but `cmd | tee f` reports tee's status, so the recipe would have sailed past exactly the failures meant to stop it. Added `set -o pipefail` with the reason stated.

**2. Archive space** — correct. Confirmed the figure from the checked-in metrics (`output_bytes: 29418788410`, ~29 GB per run), so a warm-up plus two measured sweeps need ~88 GB on top of the per-worker 65 GiB floor, and archiving onto the filesystem being measured frees nothing between workers — the runner says so itself at `run_vepyr_worker_scaling.py:242-244`. The sweep now drops the VCF once archived, since the gate reads only `summary.tsv` and the traces, and the note points at `--require-separate-filesystems` for machines with a second volume.

**3. Phase thresholds instead of an exact diff** — correct on both halves. A byte diff cannot express a 5% allowance, and `grep -h` merged the per-worker files so a regression at 8 workers could cancel against a speed-up at 1. Replaced with a bundled `scripts/compare_traces.py` that sums each `(worker, stage, event, metric)`, skips sub-50 ms phases where relative swings are meaningless, fails past the threshold, treats a phase present on one side only as a shape change rather than noise, and exits nonzero rather than passing when either side has no trace lines. Verified against synthetic traces: it flags a 20% regression at one worker count while passing 2% at the other, and refuses to pass on an empty trace.

**4. Stale downstream pins** — correct, and the subtlest of the set. A review fix landing on formats moves its head while functions still pins the step-4 commit and vepyr the old functions head, so all three PRs can go green while the measured stack excludes the accepted fixes. Step 6 now re-pins the cascade in dependency order after every upstream push, and step 7 checks the pins before measuring.

**5. Preserve field reports before the parity gate** — already fixed, in 15ed212, one commit before the review ran. The recipe now runs field mode, gates on it, then takes the digests, and copies the reports aside because the md5 pass overwrites them. Re-reported because the surrounding hunk moved; no change needed.

@codex review
@claude review

mwiewior:
Round three — all three correct, fixed in a003272. Same class as the last round: the recipe carrying on and reporting success after a step had already failed.

**Cleanup ran too late.** Confirmed against `run_vepyr_worker_scaling.py:266-272` — the free-space check runs before *every* worker, so `--workers 8 1` archived ~29 GB for worker 8 and then failed worker 1 on the check that archive had just invalidated. The sweep now calls the runner once per worker and deletes the VCF between invocations, which is what makes the cleanup actually recover space in time.

**`sweep` returned `rm`'s status.** Correct — a failed runner looked like a successful sweep. It now saves the runner status, cleans up regardless, breaks on failure and returns the saved status. Verified both directions against a stubbed runner: a failing worker propagates its exit code (7), a clean sweep returns 0 and leaves no VCF behind.

**`pipefail` without `errexit`.** Correct, and I had stopped one step short last round: pipefail made the failure visible in the pipeline status but did not stop the block, and the trailing `cp` then supplied a zero exit for the whole thing. Each pipeline now carries an explicit `|| exit 1`, with a note that the block is meant to be run as a script.

@codex review

mwiewior:
Self-audit while waiting for the next review pass. Three rounds produced nine findings and every one was a false-green path, so I grepped the recipe for commands whose exit status is never inspected. Three more of the same kind, fixed in 473c9a0.

**The wall and RSS bars were not enforced by anything.** They were stated as gates in the table, but the recipe printed a joined table and left a human to compare columns. That is how a regression gets waved through, and it is the same mistake as the earlier ones wearing different clothes.

**`compare_traces.py` was invoked without its exit status being used**, so even a detected phase regression would not have stopped the workflow.

**The step-1 trace check printed `grep -c` counts without asserting**, so a run with tracing accidentally disabled looked identical to one with traces.

`compare_traces.py` is now `compare_runs.py` and covers all three metrics in one verdict — phases from the per-worker traces, wall and RSS from `summary.tsv` — ending on a `VERDICT:` line with exit 0 or 1. Two deliberate refusals: a phase present on only one side fails as a shape change rather than noise, and an empty `annotation_seconds` is an error rather than a zero, because it means the run did not report. Verified on synthetic archives across all four paths: phase regression fails, RSS-only regression fails, an all-clean pair passes, and an empty field errors out.

One note against myself, since it is the same bug class: my first attempt to extend that script silently edited nothing. ruff-format had already reflowed the anchor text, so the string replace matched zero times and still reported success. Caught it because the new section did not appear in the output. Edits now assert their match count.

@codex review

mwiewior:
Round four — both correct, fixed in 92e2160.

**Unguarded sweep callers.** Correct, and a fair hit: the previous commit made `sweep` propagate the runner's status and then left both call sites discarding it, so a failed warm-up was followed by the measured sweep and a successful measured run returned 0 for the whole block. Fixing the function without its callers fixed nothing at the level that matters. Both calls now carry `|| exit 1`.

**Relative comparator path.** Correct — step 1's quality section leaves the shell in `$ROOT/e2e-testing/scripts`, so the gate would have failed to locate its own script before comparing anything. Now resolved from `$ROOT`. (The script is `compare_runs.py` as of 473c9a0; it absorbed the wall and RSS checks so the three bars return one verdict.)

I then re-audited every bash block for both classes rather than just these two instances: no remaining relative path outside a `cd`, and no remaining invocation whose exit status is discarded.

Tally so far, since it says something about where the risk in this file lives: eleven findings across four rounds plus my own audit, and every one was a false-green path rather than a crash. None was a wrong command; all were commands that could fail without the workflow noticing.

@codex review

mwiewior:
Round five — all six correct, fixed in 3f2016b.

**One `<n>` across three repos.** Correct, and it recurred in four loops. PR numbers differ per repo, so those loops would have commented on unrelated PRs, and the handoff loop could have marked the wrong one ready. Step 4 now records `repo:number` pairs as it creates each PR, and every later loop reads that list. Kept to `${e%%:*}` / `${e##*:}` word-splitting rather than an associative array so it still works on the bash 3.2 that ships with macOS.

**Unguarded review requests.** Correct — a failed `@codex review` followed by a successful `@claude review` returned zero and walked into the "both reviewers" gate with one of them never having looked. Guarded.

**Unguarded handoff.** Correct, and the consequence is worse than a plain failure: a hand-off that half succeeds leaves one PR a draft while its own comment claims the stack is ready, and nobody goes looking. Both commands guarded.

**PR-body read-back masked by `head`.** Correct. The verification step could report success without having verified anything, which is a poor property for a read-back whose only job is to catch `gh pr edit` silently doing nothing. pipefail plus explicit guards.

**Only `@claude` re-requested after a push.** Correct and the most embarrassing of the set: the skill states three paragraphs earlier that codex has no synchronize-triggered workflow, then re-requested only claude, guaranteeing codex never evaluated the commits actually being proposed. Both are now re-requested.

**`VERDICT: PASS` on header-only summaries.** Correct, and the one I most wanted to know about. `read_summary` returned an empty mapping, `compare_summary` iterated nothing, zero failures were counted, and the wall and RSS bars were skipped while the script printed PASS — the exact failure the script exists to prevent, inside the script itself. An empty summary is now a hard error and a comparison with no worker rows counts as a failure. Verified both directions.

Running tally: seventeen findings across five rounds plus my own audit, and every one is a false green rather than a wrong command. That is now a strong enough signal that I would treat "does this step fail loudly when its input is missing or its child errors" as the primary review question for this file, ahead of whether the commands are right.

@codex review

mwiewior:
Round six — all six correct, fixed in 242b301. Two were regressions I introduced in round five.

**Literal `<n>` in the monitor and both pin lookups.** Correct, and worse than a placeholder left behind: to the shell `<n>` is input redirection from a file named `n`, so the query fails and the trailing `sort` reports success. All three now read the recorded `repo:number` list.

**Monitor swallowed query failures.** Correct. With `2>/dev/null` and only `sort`'s status, a monitor that had lost authentication was indistinguishable from a quiet green stack — silent, forever. Each query is now checked and announces the failure.

**Re-request before the pin cascade.** Correct, and my own regression from round five: I moved the re-request loop in and placed it ahead of the pin commits, so both reviewers were asked to look at heads that were about to be replaced. Moved after the cascade.

**`cargo fmt && cargo clippy` followed by the lint.** Correct — a clippy failure was overwritten by the lint's clean exit, so PR creation could proceed with the Rust side red. Each guarded.

**Read-back compared nothing.** Correct, and it defeated the purpose: the check exists to catch a PATCH that silently kept the old body, and printing five lines proves only that the fetch worked. It now diffs the whole fetched body against the file that was sent.

**Comparator passed when both sides omitted the same worker.** Correct, and the same shape as the header-only case from round five — absence on both sides leaves no missing key to notice. The expected worker set is now stated explicitly (`--expect-workers`, default `1,8`) and checked against both traces and both summaries. Verified: a pair missing worker 1 on both sides fails, a complete pair passes.

Both regressions came from fixing a finding at one level and not re-checking the surrounding block, which is the same mistake as the `sweep` callers in round four. I have started re-auditing the whole file for the class after each fix rather than only patching the reported line.

@codex review

mwiewior:
Round seven — both correct, fixed in c7850ac. I verified each against this PR before changing anything.

**`gh pr checks` exit codes.** Correct. Measured on this PR just now: exit 8 with 7 checks pending, and `gh pr checks --help` documents `8: Checks pending`. So its status describes the checks rather than the query, and the guard I added last round would have printed `QUERY FAILED` through the entire normal review lifecycle — and stayed there for a red check, which is precisely when the monitor matters most. Replaced with `--json statusCheckRollup`, measured exit 0 under the same pending state.

**Issue comments versus review comments.** Correct, and this is the one that mattered. On this PR: 13 issue comments against 27 review comments. Every codex finding in all seven rounds has been a review comment, so a monitor polling only `gh pr view --json comments` would have reached a green snapshot with all of them unseen, while the skill's own definition of green requires each finding addressed. Now polls `repos/{owner}/{repo}/pulls/{n}/comments` alongside the issue thread.

Smoke-tested the loop body against this PR: 7 check results, 27 findings.

The irony is not lost on me — the monitor I am running for this PR polls the review-comment endpoint, which is why these rounds reached me at all, while the version I wrote into the skill polled the wrong one. The working thing and the documented thing had drifted apart, and only the documented one was under review.

@codex review

mwiewior:
Round eight — correct, fixed in ef9450e.

Measured before changing anything: `per_page=5` returns 5 of 28 without `--paginate` and all 28 with it. The default page is 30 and this PR is at 28 review comments, so the monitor was two comments away from silently reading one page — and the items that fall off the end are the newest, which are precisely the findings still outstanding. Both comment endpoints now paginate. Worth noting the per-item `--jq` filters concatenate cleanly across pages, whereas a `length` filter would not, since it emits one value per page.

The same bug was live in the monitor I am running for this PR, which counted findings with `--jq length` and would have stopped noticing new ones at 30. Restarted it with paginated id counting.

That is the second time a defect has been present in both the documented recipe and the thing I was actually running — after the review-comment endpoint last round. The two keep drifting because only one of them is under review. I have taken to smoke-testing each monitor change against this PR before committing it, which is how the page cap got measured rather than assumed.

@codex review

mwiewior:
@claude review

mwiewior:
Round nine — correct, fixed in c252ada.

Verified on this PR before changing anything: every codex pass is a review submission, and the review at `0923622` carries no inline comment at its commit. So a reviewer with nothing to say leaves a review and no comments, and the monitor saw silence — indistinguishable from a reviewer that never ran. That directly defeated step 6's gate, which asks whether both reviewers examined the current head.

The monitor now polls `pulls/{n}/reviews` as a third endpoint, and includes the reviewed commit in the emitted line so "codex has looked at *this* head" is observable rather than assumed. Smoke-tested live.

Three endpoints, three different things: review comments hold inline findings, issue comments hold the bot's status posts, and reviews hold the pass itself. Missing any one of them makes the monitor confidently wrong in a different way.

@codex review

mwiewior:
Round ten — correct, fixed in 156f5e3.

Measured on this PR: every codex review is `COMMENTED` with a 621-character body, so the state alone genuinely does not discriminate — a review carrying a finding and a clean pass project to the identical line, and dropping the id left no way to go and read the difference. The projection now emits `id` and `body_chars`, and the surrounding text shows the one-liner to fetch a body by id.

This is the third consecutive round on the same block, each one a layer further in: first it polled the wrong endpoint entirely, then it missed reviews without inline comments, now it flattened reviews that differ only in their body. Each fix made the monitor more truthful and exposed the next place it was still guessing. I have no more layers in mind for it, which given the last three rounds is weak evidence rather than strong.

@codex review

mwiewior:
Round eleven — both correct, fixed in cccf3aa.

**`gh pr create` head branch.** Correct, and confirmed in the local help text: `--head` defaults to the current branch while `--repo` only picks the target. So the loop would have proposed this checkout's branch to the other two repositories. It now passes `--head` explicitly, with the convention that all three repos use the same already-pushed branch name.

**Review-body lookup.** Correct on both counts. It referenced `$repo` and `$n`, which exist only inside the monitor's loop — and since the monitor occupies a shell, the lookup is by definition run somewhere else. Each review line now carries its own `repo#number`, and the lookup addresses `pulls/{n}/reviews/{id}` directly, which is a single object and so needs no pagination. Verified that endpoint returns the body.

One thing worth reporting against myself: while fixing this I wrote `BRANCH=fix/<slug>`, and the `bash -n` check over every fenced block rejected it — an unquoted `<slug>` is input redirection, exactly the bug you reported for `<n>` two rounds ago. I have added a sweep for unquoted placeholders across all blocks, which now passes clean. That check existing is the only reason it did not ship.

@codex review

mwiewior:
Round twelve — correct, fixed in 220dd2a, and inspecting it turned up a sharper bug underneath.

**The reported problem.** Correct: the rollup query dropped every check with an empty conclusion, so a check still running was indistinguishable from one that does not exist. The monitor now emits pending states.

**The gate you noted was missing.** You were right that nothing later verified all checks concluded. Step 8 now proves every check is green before any PR leaves draft, and refuses on a pending one. The monitor is informational; the hand-off needed its own gate, and "the others went green" is precisely how a stuck check gets handed off. Run against this PR in its current state, it correctly refuses and names all seven running checks.

**The sharper bug.** Rather than assume the shape, I dumped a raw pending entry: a `CheckRun` carries `conclusion: ""` and `status: "IN_PROGRESS"`. jq's `//` falls back only on `null` or `false`, so `.conclusion // .status` returns the empty string and loses the state entirely — my previous "fix" would have reported every pending check as blank even after I stopped filtering them out. Both call sites now test for empty explicitly. Measured: the broken form gives 7 empty states, the corrected one gives 7 `IN_PROGRESS`.

That one is worth generalising. `//` on a field that can be an empty string rather than null is a trap I should assume is present anywhere I have written a fallback against this API.

@codex review

mwiewior:
Round thirteen — correct, fixed in b6f0c7d.

`claude-code-review.yml` does list `ready_for_review` among its trigger events, so `gh pr ready` starts a fresh review check and the gate I added last round ran strictly before it. It therefore said nothing about the one check the hand-off itself creates, and the comment could have gone out while that check was running or after it failed.

Step 8 is now four ordered phases with the check pulled out into a `gate` function: refuse to leave draft while anything is red or running, mark ready, wait out the checks that marking ready started and gate again, and only then post the hand-off comment. `gate` returns 1 for "not yet" and 2 for "the query failed", which the wait loop needs so a broken credential does not spin silently for half an hour. Verified against this PR in its current state: it refuses and names all seven running checks.

The pattern across the last several rounds is that each gate I add is correct about the state at the moment it runs and wrong about the state it causes. This one is the same shape as the pin cascade in round six, where reviewers were asked before the commits that changed what they would review. Acting and then verifying the result of the action, rather than verifying and then acting, seems to be the reliable ordering here.

@codex review

mwiewior:
Round fourteen — all three correct, fixed in 445fd42. This block has now drawn three consecutive rounds and each fix made it longer and less checkable as inline shell, so I extracted it to `scripts/handoff.sh` and exercised it against this PR before committing.

**A green check is not the same as no findings.** Correct. The gate looked only at `statusCheckRollup`, and the review workflow concludes successfully regardless of whether its review carries feedback — so it could not answer step 6's actual requirement. The script now snapshots bot comment and review ids before marking ready and compares after; new feedback is a refusal, not a note.

**Stale-green rollup after `gh pr ready`.** Correct, and a subtle one: GitHub takes a moment to register the triggered run, so the first poll could accept the *previous* green. The script records the check-run set beforehand and waits for it to change before treating any green as meaningful, then fails if no post-ready run ever appears.

**Return 2 treated as pending.** Correct, and entirely my own: I introduced that return code last round precisely so a query failure could be distinguished, then wrote `gate ... && break`, which treats 1 and 2 alike. A broken credential would have spun silently for about thirty minutes. It now aborts immediately with the diagnostic printed.

Verified before pushing: phase 1 refuses against this PR in its current state, names the running checks, and stops before marking anything ready. Usage and unreadable-body guards exercised too.

While wiring it up I also wrote `uv run bash` in the invocation and caught it on the next pass — `uv run` is for Python entry points. That is the third defect in two rounds that I introduced while fixing a previous one, which is the argument for the extraction: the script can be executed, and everything I have actually caught in this file, I caught by running it rather than reading it.

@codex review

mwiewior:
Round fifteen — all four correct, fixed in efa647e. Note the paths moved in 0d618c0: the script is now `tools/vepyr-fix/handoff.sh` and the runbook `docs/runbooks/applying-a-vepyr-fix.md`, so these anchors have shifted.

**Bare repository names.** Correct, and a hard breakage: `gh --repo` requires OWNER/REPO, so phase 1 aborted every time and no PR could ever be marked ready. My earlier test passed only because I invoked the script by hand with `biodatageeks/vepyr:100` — an owner-qualified entry, a different input shape from the one the runbook actually produces. Entries are now owner-qualified where they are created, every consumer takes the repository straight from the list rather than rebuilding it, and the script refuses a bare entry up front naming the reason.

**Clean reviews counted as feedback.** Correct, and it inverted the gate: a reviewer with nothing to say still submits a review, and codex posts a fixed summary body either way, so phase 4 would have refused every clean run and the happy path could never complete. Only inline review comments now count as findings. New reviews are printed for visibility but do not fail. The residual gap — a bot putting a finding only in a review body — is documented at the call site rather than pretended away.

**Brace-group exit status.** Correct: in `{ a; b; } | sort -u` the group reports b's status and sort succeeds regardless, so a failed first query passed silently as an empty snapshot. Each query is now captured and checked separately.

**Snapshot key collisions.** Correct. Two repos can share a PR number, and keying on the number alone let the later one overwrite the earlier one's signature and feedback. Keys now include a filesystem-safe repository id.

The first one is the useful lesson. I tested the script, but with input I typed rather than input the runbook generates, so the test agreed with my assumption instead of checking it. I have re-verified this round with the documented shape end to end.

Separately, and against myself: while reading this round my own watch loop announced "all checks concluded" for the new head while twelve were queued. That is the identical stale-green race you reported for the hand-off in round fourteen — the rollup had not yet registered the new push's checks, so it read the previous commit's green. The script now guards against it; my ad-hoc loop did not.

@codex review

mwiewior:
Round sixteen — both correct, fixed in a12a637. Both were caused by the previous round's fixes.

**`eval` with a hyphenated key.** Correct, and I confirmed the mechanics: `eval "sig_biodatageeks_datafusion-bio-formats_41=hello"` exits 127 with `command not found`, and the later read returns `-bio-formats_41` — a bogus `before` that makes phase 3 see the rollup as already changed and accept the stale pre-ready green. The eval is gone entirely; the signature is now a file keyed like the other snapshots, and the key encodes every non-identifier character. A repository name is not a shell identifier and should never have been used as one.

**`num` against owner-qualified entries.** Correct. Making PRS owner-qualified last round broke the matcher, so both head lookups in the pin cascade got an empty selector — and `gh pr view` with no selector falls back to the current branch, so it would have inspected an unrelated PR rather than failing outright. It now matches either form and reports an unresolved name instead of returning empty.

Both were masked by the same testing mistake as round fifteen: I exercised the script with `biodatageeks/vepyr:100`, whose repository name contains no hyphen, so the eval succeeded and the defect stayed invisible. Testing with the one entry that happens not to trigger the bug is not testing. This round I verified against a hyphenated entry and checked `num` against both forms.

The pattern across the last three rounds is that each fix introduces its own defect in the code it touches, and my verification keeps using the convenient input rather than the representative one. The two repositories with hyphenated names are the normal case here; `vepyr` is the exception I kept reaching for.

@codex review

mwiewior:
Round seventeen — both correct, fixed in c09a865, but the fix this round is mostly a test harness.

Three consecutive rounds have now put a defect in this script that I introduced while fixing the previous one. The reason is structural: phases 2 to 5 mutate a real pull request, so I could not exercise them, and every defect in them survived until you found it. Patching a fourth time without changing that was not going to converge.

So `tools/vepyr-fix/tests/` now holds a scripted stand-in for `gh` that serves canned rollup, comment and review JSON and records mutating calls, plus a suite driving all five phases through it. 18 assertions, no network, no PR touched.

It earned its keep before it verified anything: two of my own harness bugs surfaced first. `grep -c … || echo 0` printed zero twice, and the stub advanced state on reads rather than on `gh pr ready`, which let a phase-2 snapshot already contain the finding phase 4 is supposed to notice. Both were fixed before I trusted a single verdict.

**Review-body findings.** Correct — the phase printed the id and proceeded while claiming to refuse all new feedback. Bodies are now snapshotted base64-encoded, and a review whose body has not been seen before on this PR is a refusal. Note empty-versus-non-empty cannot be the test: your reviews carry a fixed 621-character summary body on every pass, including clean ones, so novelty is the only usable signal.

**Rerun after refused feedback.** Correct, and it made the refusal path a dead end: phase 2 had already taken every PR out of draft, and marking an already-ready PR ready causes no transition, so phase 3 waited for a run that would never appear. Draft state is now recorded before marking ready, and an already-ready PR skips the wait while still requiring green.

Both are covered by tests that were red before the fix and green after.

@codex review

mwiewior:
Round eighteen — both correct, fixed in 5afbd12.

**Repeated finding bodies.** Correct, and the flaw was in my choice of discriminator. Matching against every body seen before on the PR exempts a reviewer that repeats an earlier finding verbatim, because the text is already in the history even though the review id is new. The check now works on identity as you suggested: every review new by id has its body inspected, and only the routine summary template is exempt — a fixed marker, overridable via `HANDOFF_CLEAN_BODY_MARKER`. Anything else in a new review body refuses.

**The stub could not model pending-then-green after ready.** Correct, and this is the more valuable of the two: reads never advanced the scenario, so phase 3's polling loop had never been executed by any test — the phase that three separate findings have already landed on. Reads now advance the scenario after ready, and a new test asserts the run genuinely waits through pending and reaches green rather than passing because everything was green on the first look.

Two fixture bugs of my own surfaced while doing this. My review fixtures carried a body of "x", which under the new rule is a finding, so the happy path went red — the fixture was unrealistic, not the script, since routine reviews carry the template. And the earlier body-history test had been passing for the wrong reason.

25 assertions, all green. The live path is unchanged: still refuses against this PR while its checks run.

Worth noting what the harness changed. Before it, every defect in phases 2 to 5 reached you. In the two rounds since, it has caught four problems before you saw them — two fixture bugs and two harness bugs — and the remaining findings have been about the model being too crude rather than the code being broken.

@codex review

mwiewior:
Round nineteen — correct, fixed in b082efe.

Containment was too weak exactly as you describe: a body carrying the marker alongside a finding was exempted whole.

The interesting part is what happened next. Replacing containment with "matches a body seen before" fixed your case and immediately regressed round eighteen's — a finding repeated verbatim is already in the history, so it matched and was waved through again. Each test alone lets one real case through. Both are now required: a new review body is clean only if it carries the marker **and** matches, in full, a body this PR has already shown, with commit shas and timestamps normalised out so the same template on a later commit still matches.

Four cases are now pinned by tests:

| body | verdict |
|---|---|
| clean template repeated | hand off |
| template, volatile parts changed | hand off |
| finding repeated verbatim | refuse |
| marker with a finding appended | refuse |

29 assertions, all green. I would not have noticed the regression without the round-eighteen test still being there — it went red the moment I made the change, in the same run.

@codex review

mwiewior:
Round twenty — correct, fixed in cd4d06a.

Two bugs compounding, as you describe. Storing normalised bodies in a file makes every *line* a record, and `grep -F` treats a multiline pattern as a set of alternatives, so the marker line of a multiline template matched a later body that merely began the same way. Bodies are now reduced to a sha256 of their normalised form — one line per body regardless of its content — so a whole-body comparison stays one.

The reason my round-nineteen tests missed this is worth stating: my fixture template was a single line, and the real reviewer's summary is not. The test was green because the fixture avoided the shape that breaks it. That is the third time in this loop that my verification used a convenient input rather than a representative one, after `vepyr` having no hyphen and the owner-qualified entry I typed by hand.

Two tests added using a multiline template: a finding appended below it refuses, the same template repeated hands off. 33 assertions, all green, live path unchanged.

@codex review

PR #107, opened 2026-09-12
chore: pin the fjall/Lance-scaffolding removal and drop the last legacy tokens

mwiewior:
## What

Pins the engine refactor that removes the fjall/Lance-era scaffolding from the variation lookup, and clears vepyr's own last legacy tokens.

Pin: `datafusion-bio-function-vep` moves to `e031755`, the head of biodatageeks/datafusion-bio-functions#248.

### The backend token, which must move with the pin

`src/annotate.rs` passed the literal `"lance"` as the annotation-store backend token at both call sites. The engine accepted it as a deprecated alias and resolved it to Parquet, so it was inert; #248 removes the alias. Because the token is a string validated at run time, not a type, leaving it would compile fine and fail at run time. It becomes `"parquet"` here, in the same tree as the pin.

The related `cache_format` axis needed no change: vepyr already rejects anything but `"parquet"` in both its Rust and Python guards.

### The permanent `None`

`build_cache()` returned `(entity, parquet_files, Option<...>)`, and the third element has been a hard-coded `None` since the fjall backend was removed. Dropped, with the three-name unpacks in `__init__.py` and in `tests/test_build_cache.py`, and the `.pyi` return annotations. This is the one user-visible API change: anyone unpacking three names from `build_cache()` gets two now.

`CLAUDE.md` no longer says the package converts caches to "Parquet and fjall (embedded KV store) formats".

Untouched on purpose: the tests that assert fjall artefacts are absent or that `cache_format="fjall"` is rejected. They still describe something true.

## Verification

| Gate | Result |
|---|---|
| `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings` | pass |
| `uv run pytest` | 1533 passed, 2 skipped |
| chr22 strict md5 vs Ensembl VEP 116 merged | body digest identical to the pre-change baseline |

The md5 baseline was taken on `master` before any edit and re-run on this branch with the same command: 50,861 records, strict mode, body digest `d5f1a8c3749b5f20656d845c8310380f` on both sides.

No performance sweep: the engine change is deletions and renames plus one fewer call to an idempotent guard, and this host has 34 GiB free against the ~65 GiB the whole-genome harness needs.

## Merge order

biodatageeks/datafusion-bio-functions#248 first, then re-point the pin from the PR head to the merge commit, then this.

https://claude.ai/code/session_0112cyTLXZanqcnAkVxQiZwy

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -10,7 +10,7 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 - **`datafusion-bio-format-ensembl-cache`** — reads Ensembl VEP offline cache directories (Storable/Sereal `.gz` files) into DataFusion `TableProvider`s with Arrow schemas for variations, transcripts, exons, translations, regulatory/motif features
 
 The Python package exposes:
-1. Ensembl cache conversion to Parquet and fjall (embedded KV store) formats
+1. Ensembl cache conversion to a point-lookup Parquet cache
 2. Variant annotation against converted caches
 
 ## Build & Development
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1273,9 +1273,8 @@ dependencies = [
 [[package]]
 name = "datafusion-bio-function-vep"
 version = "0.19.0"
-source = "git+https://github.com/biodatageeks/datafusion-bio-functions.git?rev=9f74fd9caa0682d4706f0d0f7e83bfa38eface94#9f74fd9caa0682d4706f0d0f7e83bfa38eface94"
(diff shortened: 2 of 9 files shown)


mwiewior:
@codex review

mwiewior:
@claude review

mwiewior:
@codex review

mwiewior:
@claude review

mwiewior:
@codex review

mwiewior:
@claude review

PR #108, opened 2026-09-12
docs: variation-lookup figure and the script that measures it

mwiewior:
## What

A manuscript figure for the variation-lookup path, and the two scripts that produce it. The scripts are the point: every quantity on the figure can be re-derived rather than taken on trust.

`docs/diagrams/` already pairs a source with its rendered SVG, so these sit alongside the existing mermaid diagrams.

| File | Role |
|---|---|
| `variation-lookup.py` | draws the figure; all quantities in one `MEASURED` dict at the top |
| `variation-lookup-measure.py` | produces those quantities from a real cache and a real VCF |
| `variation-lookup.svg` | the rendered figure |

## The figure

Four panels: an input buffer reduced to probe positions; why the shard is clustered by tier; the two reads that resolve the probes; and the join back to the buffer.

The cell strips in panel B are schematic and are labelled as such on the figure. Everything else, including the tier threshold, the page geometry and the whole measured block, is real.

## What the measurement does

It replays one real 5,000-record chr22 buffer against a real shard twice, once in the shipped `(tier, position)` order and once against a position-only counterfactual built from the same rows and the same page row-counts, so the two differ by clustering alone. It reports what the locate read costs: the pages it opens and the rows it decodes.

For the chr22 shard of the GRCh38 merged cache for Ensembl VEP 116, probed with the HG002 benchmark VCF, matching 6,178 cached rows:

| Row layout | Pages read of 15,381 | Position rows scanned | Matches per page |
|---|---|---|---|
| Sorted by position only | 920 | 928,028 | 6.7 |
| As shipped, tier 0 first | 220 | 207,732 | 28.1 |

The split behind that average: 96% of the matches are common variants served by 83 tier-0 pages at 71.6 rows each, while the rare 4% still scatter over 137 tier-1 pages at 1.7 rows each.

## Reproducing

```
python docs/diagrams/variation-lookup.py            # redraw from MEASURED
python docs/diagrams/variation-lookup-measure.py \
    --shard  $DATA_VEPYR_DIR/cache/116_GRCh38_merged/variation/chr22.parquet \
    --pages  chr22_pages.csv \
    --vcf    $DATA_VEPYR_DIR/HG002_GRCh38_1_22_v4.2.1_benchmark.vcf.gz
```

pyarrow cannot read the Parquet offset index, so the page boundaries come from a short Rust program built against the same `parquet` crate the engine uses. It is carried verbatim in the measure script's docstring rather than left as an undocumented step.

## Scope

Docs only. No library code, no tests, no pins.

https://claude.ai/code/session_0112cyTLXZanqcnAkVxQiZwy

### DIFF
--- a/docs/diagrams/variation-lookup-measure.py
+++ b/docs/diagrams/variation-lookup-measure.py
@@ -1,0 +1,173 @@
+#!/usr/bin/env python3
+"""Measure every quantity that appears in ``variation-lookup.py``'s MEASURED table.
+
+The figure claims that clustering the variation shard by tier, rather than by
+position alone, makes the lookup's locate read touch far less data. This script
+is where that claim comes from. It replays one real annotation buffer against a
+real shard twice: once in the order the shard actually ships, and once against a
+position-only counterfactual built from the same rows and the same page sizes.
+
+Nothing here is estimated. Run it, then paste the printed values into MEASURED.
+
+Inputs
+------
+--shard      one per-chromosome variation Parquet file from a built cache
+--pages      CSV of the position column's page boundaries, see below
+--vcf        a bgzipped, tabix-indexed VCF to draw the probe buffer from
+--chrom      contig name as it appears in the VCF
+--buffer     records in the buffer (5000 matches the engine's lookup slice)
+
+Producing --pages
+-----------------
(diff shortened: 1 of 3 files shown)


mwiewior:
@codex review

mwiewior:
@claude review

PR #111, opened 2026-09-12
ci: build and test Linux aarch64 wheels on native arm runners

mwiewior:
## Summary

Adds Linux arm64 (aarch64) support.

- **Wheels:** `aarch64` added to the manylinux matrix in `ci.yml` and `publish_to_pypi.yml`. Built natively on `ubuntu-24.04-arm` runners, not cross-compiled. `publish` already collects every `wheels-*` artifact, so it is unchanged.
- **Tests:** new `linux-arm64-tests` job runs the pytest suite on `ubuntu-24.04-arm` with Python 3.12 (one interpreter, since the wheel is abi3, same reasoning as the Windows job).
- **bioconda:** `linux-aarch64` added to `additional-platforms`.
- **Docs:** platform lists in `docs/index.md` and `docs/developers.md`.

- **Dev dependencies:** `polars-bio` removed from the `dev` group. Nothing in the suite called it (`test_sink_vcf` only imported it to skip when it was missing), and it has no Linux aarch64 wheel, so `uv sync` on the arm64 job would have compiled it from source. The notebook cell that reads the golden VCF with it now notes the separate install.

No Rust changes. The one arch-specific dependency on the annotation path is `coitrees` 0.4, which selects its NEON interval tree at compile time on aarch64 (NEON is a default target feature there); the macOS arm64 wheel already ships that path, and `linux-arm64-tests` exercises it end to end. vepyr's own `cargo test` has no arch-specific code and does not reach it.

## Motivation

A Seqera Wave build of `pip install vepyr==0.7.0` for `linux/arm64` fails: PyPI has no aarch64 wheel, pip falls back to the sdist, and the image has no C toolchain (`error: linker cc not found`). Build: https://wave.seqera.io/view/builds/bd-f73652bf66f6062f_1

## Test plan

- [ ] `linux (aarch64)` wheel job builds a `manylinux_*_aarch64` abi3 wheel
- [ ] `linux-arm64-tests` passes (LFS fixtures, bcftools/tabix via apt, golden outputs)
- [ ] After release: rerun the Wave build for `linux/arm64`

https://claude.ai/code/session_01JD5mVRErcEzPWfHZtQxWhK

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -103,11 +103,37 @@ jobs:
           uv sync --python ${{ matrix.python-version }}
           uv run pytest -v
 
+  # Linux arm64 on a native GitHub arm runner, not cross-compiled, so the
+  # same host that builds the wheel also runs the suite. One interpreter for
+  # the same abi3 reasoning as the Windows job above.
+  linux-arm64-tests:
+    runs-on: ubuntu-24.04-arm
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ["3.12"]
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          lfs: true
+      - uses: dtolnay/rust-toolchain@stable
+      - uses: Swatinem/rust-cache@v2
+      - uses: astral-sh/setup-uv@v5
+      - name: Install VCF command-line tools
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y bcftools tabix
+      - name: Build and test
+        run: |
+          uv python install ${{ matrix.python-version }}
+          uv sync --python ${{ matrix.python-version }}
+          uv run pytest -v
+
   linux:
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.target == 'aarch64' && 'ubuntu-24.04-arm' || 'ubuntu-latest' }}
     strategy:
       matrix:
-        target: [x86_64]
+        target: [x86_64, aarch64]
     steps:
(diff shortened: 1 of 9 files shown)


mwiewior:
@codex review

mwiewior:
@claude review

mwiewior:
Re the Claude review above:

**1. `test_sink_vcf`: renamed, no coverage was lost.** On master the test never called `sink_vcf`. It imported `polars_bio` only to decide whether to skip, then wrote a TSV with `df.write_csv(..., separator="\t")`, so it never checked VCF-formatted output. The name was misleading, though: f1d247c renames it to `test_collect_to_tsv` and changes the temp-file suffix to `.tsv`. VCF output is covered by `test_annotate_to_vcf_output`, `test_annotate_vcf_has_csq_header`, the `*_forwards_to_vcf_writer` tests and the golden suite.

**2. aarch64 `cargo test` leg: not adding it.** vepyr's `src/` has 4 Rust tests (all in `annotate.rs`) and no arch-specific code, and `cargo test` does not run the unit tests of dependency crates. The arch-specific code on the annotation path is `coitrees` 0.4, which picks its NEON tree at compile time on aarch64. vepyr's `cargo test` never reaches it, whereas `linux-arm64-tests` runs it end to end through the Python suite, and the macOS arm64 wheel already ships the same path.

Checking this also turned up a wrong claim in the PR description: the `neon` code it cited lives in `bio-function-ranges`, which vepyr doesn't depend on. The description now refers to `coitrees` instead.

